PR #135, opened 2014-05-13
Auto-update feature

ioannis-e:
- [x] download setup file from www.sourceforge.net depending on architecture
- [x] verify embedded signature
- [x] run setup file
- [x] close mpc-hc instance
- [x] implement update for portable instances (ie not installed) - download and save 7z archive
- [ ] decide download location
- [ ] implement updating to nightlies
- [ ] internationalization of new strings

image


XhmikosR:
@ioannis-e: care to rebase this and sort a couple of things in order to get this in?

ioannis-e:
@XhmikosR wow just to note that @Stanzilla last post was exactly 2 years ago :D

This pr was initially based on ``mpc-hc:master``, should i rebase on ``mpc-hc:master`` or ``mpc-hc:develop`` ?

If i remember correctly (i need to go through the discussion again) certain changes to the installer were previously discussed that need to be addressed as well.

XhmikosR:
Yeah, I know, things were slow. Just trying to clean up the PRs and get things going. You can close this and make a new one against develop if you find it easier.

kasper93:
> You can close this and make a new one against develop if you find it easier.

No. I changed base branch to develop. Github added this feature finally.

kasper93:
> 1. mpc-hc.org should serve links for latest stable/nightly, x86/x64 that will redirect to the actual download locations, based on input url.
eg. http://www.mpc-hc.org/download/ [release/nightly] / [x86/x64]

Fine, with additional 7z/portable variant. I think if someone haven't used installer to install MPC-HC we should ask where to download new package. Possibly with automation to extract files in current working dir.                                                                            

> 2. installer should run silently on installations as it will use the previous install selections.

Yes, and auto update should be only triggered from MPC-HC instance which matches install directory in register. Otherwise we fall into portable case.

> Points for further discussion:
> 1. where should the downloader save the file for installations? temp folder / downloads folder ?

In temp. https://msdn.microsoft.com/en-us/library/windows/desktop/aa364992.aspx

> Regarding non-installations:
> 1. Installer can be run with commandline to override installation path. This can be set to path of the running executable.

If so installer should be modified to use portable mode. So don't modify registry and stuff like that. It is possible to do without much hassle. I'd agree in portable case only 7z package is downloaded for user, but installer solution will be more convenient for fast update.

Ok. What is done and what remain to be done? Is the checklist at top valid?

Stanzilla:
Not sure if it fits mpc-hc but https://github.com/Squirrel/Squirrel.Windows is quite cool.

PR #145, opened 2014-11-01
Add IPTV capabilities to the existing Digital TV UI

xpc1000:
Extending the current interface for using it not only for DVB but also for IPTV.

Commits:
1- Add IPTV capabilities to the existing DVB UI
- Basic IPTV capabilities
- Navigation TV panel: tabbed filters for DVB-T, IPTV and Radio

2- Digital TV: adapting the channel management UI for IPTV:
- Channel Scan is now called Digital TV tools
- UI: new tab for managing IPTV channels
- IPTV: option for adding individual channels
- IPTV: option for importing a m3u list of channels
- New checkbox for removing existing channels when saving. Removes only the channels of the same type (DVB or IPTV)
- Adding channels: new counter for preventing having duplicate prefNumber channels

3- Digital TV: split options from analog capture and new Digital TV options

3.1. General
- Split menu entry into: Open Digital TV / Open Analog Capture. File->Open Device now only refers to analog device
- Split class OpenDeviceData into OpenDeviceAnalogData and OpenDeviceDigitalData
- Channel navigation panel: only enabled tabs visible (depending on Digital TV options). Set initial active tab depending on current channel type.
- Add IGMP membership management: manage IGMP group join/leave for multicast IPTV streaming sources. Supports only IPv4 and IGMPv2. IPv6 and IGMPv3 could be added in the future if needed.
- Add shortcut for Digital TV

3.2. Exception handling:
- error controls (checking devices, etc.) moved from the general entry point to each channel switch operation
- Change current behavior: when error switching to a channel -> show error message in info bar and then stay ready for switching to another channel
- If the IPTV switching channel operation fails, we don't change the window size

3.3. Options & preferences:
- Split Digital TV options from analog capture
- New Digital TV options: DVB (BDA) and IPTV (include option: enable/disable IGMP multicast membership management)

Pending issues and known limitations:
- ONLY English version: to be updated once it is ready for merge
- IPTV Channel management: additional options (scan range of IP's and service discovery for multicast) are in progress and will be included in a different PR. 
- Access to channel management from options menu: will be also included in a future PR.


src/mpc-hc/IPTVMcastTools.cpp
```diff
@@ -0,0 +1,209 @@
+/*
+* (C) 2009-2014 see Authors.txt
```
XhmikosR: Since these news files appear to be new code, just use 2014.

src/mpc-hc/IPTVMcastTools.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+{
+}
+
+CIPTVMcastTools::~CIPTVMcastTools()
+{
+}
+
+
+SOCKET CIPTVMcastTools::udp_socket_create(UINT32 local_addr, USHORT port)
```
XhmikosR: The functions' naming scheme you follow is different from what we tend to use everywhere else.

/CC @Underground78 @kasper93

src/mpc-hc/IPTVMcastTools.cpp
```diff
@@ -0,0 +1,209 @@
...
+typedef int socklen_t;
+
+#define UDP_TX_BUF_SIZE 32768
+#define UDP_MAX_PKT_SIZE 65536
+
+
+
+CIPTVMcastTools::CIPTVMcastTools()
+    : m_udp_fd(0),
+      m_socket(0),
```
XhmikosR: Please put the commas to the left in the constructor's initialization list.

src/mpc-hc/TVToolsDlg.cpp
```diff
@@ -0,0 +1,175 @@
...
+    : CDialog(CTVToolsDlg::IDD, pParent)
+    , m_TunerScanDlg(this)
+    , m_IPTVScanDlg(this)
+{
+
+}
+
+CTVToolsDlg::~CTVToolsDlg()
+{
+
```
XhmikosR: I'd remove this empty line(s) in the ctor/dtor case.

src/mpc-hc/TVToolsDlg.h
```diff
@@ -0,0 +1,62 @@
...
+class CTVToolsDlg : public CDialog
+{
+    DECLARE_DYNAMIC(CTVToolsDlg)
+
+public:
+    CTVToolsDlg(CWnd* pParent = nullptr);   // standard constructor
+    virtual ~CTVToolsDlg();
+
+    // Dialog Data
+    enum { IDD = IDD_SCAN };
```
XhmikosR: Needs `resource.h` as an include.

src/mpc-hc/AppSettings.h
```diff
@@ -549,8 +551,11 @@ class CAppSettings
     bool            fBDAIgnoreEncryptedChannels;
     int             nDVBLastChannel;
     std::vector<CDVBChannel> m_DVBChannels;
+    UINT            nNextChannelCount;
```
XhmikosR: I would name this `uNextChannelCount`.

src/mpc-hc/DVBChannel.h
```diff
@@ -96,6 +96,27 @@ struct DVBStreamInfo {
     LCID GetLCID() { return ISO6392ToLcid(CStringA(sLanguage)); };
 };
 
+// A list should be used here!!
+static bool IsValidUrl(CString strUrl)
+{
+    bool bResult = false;
```
XhmikosR: Do we even need `bResult` here?

xpc1000: It was useful for debugging, but I'll remove it now.

src/mpc-hc/IPTVMcastTools.cpp
xpc1000: OK. I'll apply that to all the new files.

src/mpc-hc/IPTVMcastTools.cpp
xpc1000: Not sure I got it. Could you be more precise?

src/mpc-hc/IPTVMcastTools.cpp
xpc1000: OK

src/mpc-hc/TVToolsDlg.cpp
xpc1000: OK

src/mpc-hc/TVToolsDlg.h
xpc1000: OK

src/mpc-hc/AppSettings.h
xpc1000: OK

src/mpc-hc/AppSettings.cpp
```diff
@@ -400,6 +403,7 @@ void CAppSettings::CreateCommands()
     ADDCMD((ID_FILE_OPENMEDIA,                  'O', FVIRTKEY | FCONTROL | FNOINVERT,         IDS_AG_OPEN_FILE));
     ADDCMD((ID_FILE_OPENDVDBD,                  'D', FVIRTKEY | FCONTROL | FNOINVERT,         IDS_AG_OPEN_DVD));
     ADDCMD((ID_FILE_OPENDEVICE,                 'V', FVIRTKEY | FCONTROL | FNOINVERT,         IDS_AG_OPEN_DEVICE));
+    ADDCMD((ID_FILE_OPENDIGITALTV,              'B', FVIRTKEY | FCONTROL | FNOINVERT, IDS_AG_OPEN_DIGITALTV));
```
kasper93: indentation

src/mpc-hc/AppSettings.cpp
```diff
@@ -1591,16 +1599,17 @@ void CAppSettings::LoadSettings()
...
-    // Save analog capture settings
-    iDefaultCaptureDevice = pApp->GetProfileInt(IDS_R_SETTINGS, IDS_RS_DEFAULT_CAPTURE, 0);
+    // Load analog capture settings
+    bEnabledAnalogCapture = !!pApp->GetProfileInt(IDS_R_CAPTURE, IDS_RS_ENABLE_ANALOGCAPTURE, FALSE);
     strAnalogVideo        = pApp->GetProfileString(IDS_R_CAPTURE, IDS_RS_VIDEO_DISP_NAME, _T("dummy"));
     strAnalogAudio        = pApp->GetProfileString(IDS_R_CAPTURE, IDS_RS_AUDIO_DISP_NAME, _T("dummy"));
     iAnalogCountry        = pApp->GetProfileInt(IDS_R_CAPTURE, IDS_RS_COUNTRY, 1);
 
+    // Load digital TV settings
+    bEnabledDVB = !!pApp->GetProfileInt(IDS_R_DVB, IDS_RS_ENABLE_DVB, FALSE);
```
kasper93: also indentation

src/mpc-hc/AppSettings.cpp
```diff
@@ -1611,6 +1620,9 @@ void CAppSettings::LoadSettings()
     nDVBRebuildFilterGraph = (DVB_RebuildFilterGraph) pApp->GetProfileInt(IDS_R_DVB, IDS_RS_DVB_REBUILD_FG, DVB_REBUILD_FG_WHEN_SWITCHING);
     nDVBStopFilterGraph = (DVB_StopFilterGraph) pApp->GetProfileInt(IDS_R_DVB, IDS_RS_DVB_STOP_FG, DVB_STOP_FG_WHEN_SWITCHING);
 
+    nNextChannelCount = pApp->GetProfileInt(IDS_R_DVB, IDS_RS_DVB_NEXTCHANNELCOUNT, 0);
+    bEnabledIPTV = !!pApp->GetProfileInt(IDS_R_DVB, IDS_RS_ENABLE_IPTV, FALSE);
+    bUseIGMPMembership = !!pApp->GetProfileInt(IDS_R_DVB, IDS_RS_USE_IGMPMEMBERSHIP, FALSE);
```
kasper93: and here ;)

src/mpc-hc/DVBChannel.h
```diff
@@ -207,3 +233,13 @@ class CDVBChannel
 
     void FromString(CString strValue);
 };
+
+static bool IsChannelIPTV(const CDVBChannel channel)
```
kasper93: const reference, we don't need copy here.

Ps. Those methods looks little bit redundant.

src/mpc-hc/DVBChannel.h
```diff
@@ -207,3 +233,13 @@ class CDVBChannel
...
+
+static bool IsChannelIPTV(const CDVBChannel channel)
+{
+    return channel.IsIPTV();
+};
+
+static bool IsChannelDVB(const CDVBChannel channel)
+{
+    return channel.IsDVB();
+};
```
kasper93: new line

src/mpc-hc/IPTVMcastTools.cpp
```diff
@@ -0,0 +1,209 @@
...
+      m_is_multicast(false)
+
+{
+}
+
+CIPTVMcastTools::~CIPTVMcastTools()
```
kasper93: empty dtor is empty, you can use default one

src/mpc-hc/IPTVMcastTools.cpp
kasper93: We use UpperCamelCase for methods name generally.

src/mpc-hc/IPTVMcastTools.cpp
```diff
@@ -0,0 +1,209 @@
...
+    int iRet = -1;
+    SOCKET udp_fd = -1;
+
+    // Initialize Winsock
+    if (WSAStartup(MAKEWORD(2, 2), &m_wsa) == 0) {
+        // Create socket
+        udp_fd = socket(AF_INET, SOCK_DGRAM, 0);
+        if (udp_fd != INVALID_SOCKET) {
+
+            memset(&m_sockaddr, 0, sizeof(m_sockaddr));
```
kasper93: `ZeroMemory()`

src/mpc-hc/IPTVMcastTools.cpp
```diff
@@ -0,0 +1,209 @@
...
+{
+    closesocket(m_udp_fd);
+    WSACleanup();
+    m_udp_fd = 0;
+    m_socket = 0;
+    m_is_multicast = false;
+}
+
+
+CString CIPTVMcastTools::get_IPv4_Addr(CString sAddr)
```
kasper93: pass CString by const reference

src/mpc-hc/IPTVMcastTools.cpp
```diff
@@ -0,0 +1,209 @@
...
+            iPos1 = sAddr.GetLength() - 1;
+        }
+        sIPAddr = sAddr.Mid(iBase + iPos0, iPos1 - iPos0 - iBase);
+    } else {
+        ASSERT(FALSE);
+    }
+    return sIPAddr;
+}
+
+USHORT CIPTVMcastTools::get_IPv4_Port(CString sAddr)
```
kasper93: pass CString by const reference

src/mpc-hc/IPTVMcastTools.cpp
```diff
@@ -0,0 +1,209 @@
...
+                WSACleanup();
+                ASSERT(FALSE);
+            }
+        }
+    }
+
+    return iResult;
+}
+
+int CIPTVMcastTools::udp_multicast_join_group(CString sAddr)
```
kasper93: here too

src/mpc-hc/IPTVMcastTools.cpp
```diff
@@ -0,0 +1,209 @@
+/*
```
kasper93, line 1: This file is in `Resource Files` folder in solution...

xpc1000, line 1: Oops :(
Fixed

src/mpc-hc/PPageDigitalTV.cpp
```diff
@@ -0,0 +1,314 @@
...
+    EnableToolTips(TRUE);
+
+    return TRUE;
+}
+
+BOOL CPPageDigitalTV::OnApply()
+{
+    UpdateData();
+    SaveFoundDevices();
+    AfxGetMainWnd()->PostMessageW(WM_DTV_REFRESHSETTINGS);
```
kasper93, line 139: You could use our event dispatcher. See `EventDispatcher.h`

alexmarsev, line 139: Actually no. This "event" can be called a command because it can't have one-many pub/sub relationship. Just leave it like this, I will refactor it after the merge.

kasper93, line 139: And here we have one-one relationship. He could even call method directly.

src/mpc-hc/IPTVMcastTools.cpp
xpc1000: OK

src/mpc-hc/AppSettings.cpp
xpc1000: OK

src/mpc-hc/AppSettings.cpp
xpc1000: OK all indentation.

src/mpc-hc/DVBChannel.h
xpc1000: The methods are used in  https://github.com/xpc1000/mpc-hc/commit/97d8f4207e93fa388168e4b8af3d8c8a56c0d1c4#diff-57f982ec5ae960ab2571cf1d6e1fa088R147: 
<code>
auto new_end = std::remove_if(std::begin(DVBChannels), std::end(DVBChannels), IsChannelIPTV);
</code>
These methods are required as the condition for the iterator and I don't see how to change them to const reference without changing the definition of std::remove_if(). Please advise.

src/mpc-hc/DVBChannel.h
xpc1000: OK

src/mpc-hc/IPTVMcastTools.cpp
xpc1000: OK

src/mpc-hc/IPTVMcastTools.cpp
xpc1000: OK

src/mpc-hc/IPTVMcastTools.cpp
xpc1000: OK

src/mpc-hc/IPTVMcastTools.cpp
xpc1000: OK

src/mpc-hc/IPTVMcastTools.cpp
xpc1000: OK

src/mpc-hc/DVBChannel.h
kasper93: If it is only used there use lambda :)         

``` cpp
auto new_end = std::remove_if(DVBChannels.begin(), DVBChannels.end(),
        [](const CDVBChannel & c) { return c.IsIPTV(); });
```

xpc1000: OK I'll try it. Thanks

src/mpc-hc/PPageDigitalTV.cpp
xpc1000, line 139: I tend to avoid calling directly methods of another window class from within a windows class if possible. I prefer using messages for modularity in this case.

BTW this message WM_DTV_REFRESHSETTINGS is being used after changing options to force the player to refresh and apply the changes automatically. It might be updated to be used for other playback modes as well. If this was the case, may be the EventDispatcher would be meaningful. 

Please let me know how to proceed. I'll do whatever is decided here.

Underground78, line 139: I think we already have something for the cases like that. I can't check
now but maybe check some other option pages.

alexmarsev, line 139: > I think we already have something for the cases like that.

We have `void CMainFrame::UpdateControlState(UpdateControlTarget target)` that we already (mis)use for such things.

> Please let me know how to proceed.

Either your current code, `CMainFrame::UpdateControlState()` or you can try using EventDispatcher.
The last one if preferred. Looking at `CMainFrame::OnRefreshPlayerSettings()`, subscribers will be `CPlayerNavigationBar`(or `CPlayerNavigationDialog`) and `CMainFrame`. Publisher will be `CPPageDigitalTV` obviously. For how, search for `MpcEvent::` in the code, there are plenty of examples. Just note that while the event dispatcher itself is synchronous, the order subscribers get notified of the same event is undefined.

src/mpc-hc/DVBChannel.h
```diff
@@ -96,6 +96,39 @@ struct DVBStreamInfo {
...
+} ();
+
+
+static bool IsValidUrl(LPCTSTR strUrl)
+{
+    auto it = find_if(Supported_Protocols.begin(), Supported_Protocols.end(),
+        [&](CString const & Protocol) { return (((CString)strUrl).Find(Protocol) == 0); });
+    return (it != Supported_Protocols.end());
+}
+
```
kasper93: If I may suggest something, I would convert string to lower instead of listing all strings. Something like this should work. :) 

``` diff
diff --git a/src/mpc-hc/DVBChannel.h b/src/mpc-hc/DVBChannel.h
index de52ac6..d5055ed 100644
--- a/src/mpc-hc/DVBChannel.h
+++ b/src/mpc-hc/DVBChannel.h
@@ -21,6 +21,7 @@
 #pragma once

 #include <vector>
+#include "ArrayUtils.h"

 #define FORMAT_VERSION_0       0
 #define FORMAT_VERSION_1       1
@@ -96,37 +97,28 @@ struct DVBStreamInfo {
     LCID GetLCID() { return ISO6392ToLcid(CStringA(sLanguage)); };
 };

-static std::vector<CString> Supported_Protocols = [] {
-    std::vector<CString> Protocols;
-    Protocols.push_back(_T("rtp://"));
-    Protocols.push_back(_T("udp://"));
-    Protocols.push_back(_T("rtsp://"));
-    Protocols.push_back(_T("mms://"));
-    Protocols.push_back(_T("http://"));
-    Protocols.push_back(_T("https://"));
-    Protocols.push_back(_T("rtspu://"));
-    Protocols.push_back(_T("rtspm://"));
-    Protocols.push_back(_T("rtspt://"));
-    Protocols.push_back(_T("rtsph://"));
-    Protocols.push_back(_T("RTP://"));
-    Protocols.push_back(_T("UDP://"));
-    Protocols.push_back(_T("RTSP://"));
-    Protocols.push_back(_T("MMS://"));
-    Protocols.push_back(_T("HTTP://"));
-    Protocols.push_back(_T("HTTPS://"));
-    Protocols.push_back(_T("RTSPU://"));
-    Protocols.push_back(_T("RTSPM://"));
-    Protocols.push_back(_T("RTSPT://"));
-    Protocols.push_back(_T("RTSPH://"));
-    return Protocols;
-} ();
-
-
-static bool IsValidUrl(LPCTSTR strUrl)
+static bool IsValidUrl(CString strUrl)
 {
-    auto it = find_if(Supported_Protocols.begin(), Supported_Protocols.end(),
-        [&](CString const & Protocol) { return (((CString)strUrl).Find(Protocol) == 0); });
-    return (it != Supported_Protocols.end());
+    auto supportedProtocols = make_array(
+                                  _T("rtp://"),
+                                  _T("udp://"),
+                                  _T("rtsp://"),
+                                  _T("mms://"),
+                                  _T("http://"),
+                                  _T("https://"),
+                                  _T("rtspu://"),
+                                  _T("rtspm://"),
+                                  _T("rtspt://"),
+                                  _T("rtsph://"),
+                                  _T("rtmp://"),
+                                  _T("rtmpt://")
+                              );
+
+    auto it = find_if(supportedProtocols.cbegin(), supportedProtocols.cend(), [&](LPCTSTR protocol) {
+        return strUrl.MakeLower().Find(protocol) == 0;
+    });
+
+    return it != supportedProtocols.cend();
 }

 class CDVBChannel

```

kasper93:
@Underground78 @alexmarsev Will you have time to test this maybe?

src/mpc-hc/mpc-hc.rc
```diff
@@ -876,58 +876,71 @@ BEGIN
...
+    LTEXT           "Receiver",IDC_STATIC6,10,75,74,8
+    COMBOBOX        IDC_COMBO3,84,73,202,30,CBS_DROPDOWNLIST | CBS_AUTOHSCROLL | WS_VSCROLL | WS_TABSTOP
+    LTEXT           "Channel switching approach:",IDC_PPAGECAPTURE_ST10,10,93,260,8
+    LTEXT           "Rebuild filter graph",IDC_PPAGECAPTURE_ST11,10,107,74,8
+    EDITTEXT        IDC_PPAGECAPTURE_DESC1,84,119,202,20,ES_MULTILINE | ES_READONLY | NOT WS_BORDER | NOT WS_TABSTOP
+    COMBOBOX        IDC_COMBO6,84,105,201,30,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    LTEXT           "Stop filter graph",IDC_PPAGECAPTURE_ST12,10,141,74,8
+    COMBOBOX        IDC_COMBO7,84,139,201,30,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    
+    GROUPBOX        "",IDC_STATIC1,5,179,298,42
```
kasper93: This groupbox is wrong positioned.

kasper93:
After adding IPTV channel with wrong URL, I `GetDefaultAudioType()` returns `-1`. Generally problem comes from fact that if few places you assume that if URL is invalid we are dealing with DVB channel, which is obviusly not correct in this case. Failing stack
```
>	mpc-hc.exe!std::_Debug_message(const wchar_t * message, const wchar_t * file, unsigned int line) Line 15	C++
 	mpc-hc.exe!std::array<DVBStreamInfo,10>::operator[](unsigned int _Pos) Line 164	C++
 	mpc-hc.exe!CDVBChannel::GetDefaultAudioType() Line 170	C++
 	mpc-hc.exe!CFGManagerBDA::CFGManagerBDA(const wchar_t * pName, IUnknown * pUnk, HWND__ * hWnd) Line 321	C++
 	mpc-hc.exe!CMainFrame::OpenCreateGraphObject(OpenMediaData * pOMD) Line 10446	C++
 	mpc-hc.exe!CMainFrame::OpenMediaPrivate(ATL::CAutoPtr<OpenMediaData> pOMD) Line 11886	C++
 	mpc-hc.exe!CGraphThread::OnOpen(unsigned int wParam, long lParam) Line 86	C++
```

xpc1000:
Thanks for your comments. I'll take a look this week.

xpc1000:
I went through your comments and fixed the things you pointed and some others. Additionally I added the following commits:

4- IPTV channel management: added service discovery  and IP Scan options
- Added service discovery option in IPTV Channels Management. To be used only for multicast. I could only test with my IPTV provider, Movistar TV. We need to test how it works with other providers. At least the tags will have to be adapted. Since the list includes all the possible channels (not only the ones that have been contracted), the process scans all the channels included and checks whether they are streaming something or not. This is reflected in the column "validated". If "Save only validated channels" checked, only the channels with validated = "yes" are saved.
- Add IP Scan range of IP's option in IPTV Channels Management. The protocol is currently rtp only. I will have to add an option for utp.
- Access to IPTV channels management dialog from both from Options menu and Digital TV.

5- DVB: Create a separate filter graph for tuner scan
- Create a filter graph separated from the main filter graph for DVB Channels Management.  When performing tuner scan from DVB channels management, a dedicated filter graph with the required filters only is created in a separate thread
- Access to DVB Channels management dialog from both options menu and Digital TV. 
- Removed DoTunerScan, StartTunerScan and StopTunerScan from MainFrm.cpp

kasper93:
Unfortunally I don't have access to any IPTV, so I can't test this part. I will look at DVB changes. I will try to do full review this weekend.

xpc1000:
And here is a link for those who want to test it:

http://www.mediafire.com/download/cfbtqzk8dyr8dqe/mpc-hc_x86.7z

Jellyfrog:
It would be cool if it could do some automatic service discovery, as defined in RFC 5328:
```
To obtain the initial Service Discovery and Selection (SD&S) XML
record, a client must by default first join the IANA registered
well-known multicast addresses 224.0.23.14 (DvbServDisc on IPv4)
and/or FF0X:0:0:0:0:0:0:12D (DvbServDisc on IPv6) and try to
obtain a boot-strap record from the IANA registered well-known
port dvbservdsc (port number 3937) via tcp and udp (see
http://www.iana.org/assignments/port-numbers).

To discover non-default entry points addresses, [TS102034] defines
that a list of Service Discovery and Selection (SD&S) entry points
addresses may be acquired via DNS according to the service
location resource record (SRV RR) [RFC2782].  The service name is
"_dvbservdsc"; the protocol may be tcp or udp, while the rest of
the name is the domain name maintained by DVB for service
discovery.  This domain name is set to "services.dvb.org".  So the
lookup shall be either "_dvbservdsc._tcp.services.dvb.org" or
"_dvbservdsc._udp.services.dvb.org".  This requires that the
terminal support an SRV cognizant DNS client and in a way
according to the specification in [RFC2782].  The DVB organization
will maintain the services.dvb.org domain name for service
discovery.  HTTP servers will be found via the tcp protocol method
whilst the multicast addresses will be found via the udp protocol
method.
```
But it's far from a blocker :)

xpc1000:
Thanks for the information. I'll take a look and try to verify in my network using Wireshark.

xpc1000:
@Jellyfrog 
I tested joining multicast address 224.0.23.14:3937 in my network and don't receive any packets. Do you receive any packet from that address? It is easy to test by entering options->DigitalTV->IPTV Channels: select services discovery and press setup. Then you can enter this address and port as entry point and press Find. You can check the packets you receive by using Wireshark for example.

LuanMax:
With this last build I still have to include my DVB-C channels under "DVB-T".

So would'nt "TV" / "IPTV" / "Radio" be better here?

The complete technical terms, imho, would be "DVB-TV" / "IPTV" / "DVB-Radio". But I'm afraid this would be too long, so "TV" / "IPTV" / "Radio" seens to me the most plausible variant (with "IPTV" compared to the DVB services "TV" and "Radio").

xpc1000:
> So would'nt "TV" / "IPTV" / "Radio" be better here?

I find "TV" too general. I prefer the long option you pointed ("DVB-TV" / "IPTV" / "DVB-Radio"). We might reduce the "DVB-Radio" to just "radio" (as it is now) to save some space. Any other opinion?

LuanMax:
Another solution could be to implement some kind of auto-detection for the broadcast transmission of DVB over Terrestrial (DVB-T), Cable (DVB-C) or Satellite (DVB-S).

Do you see any possibility here?

Underground78:
I already thought about that for other uses but there is no easy what to do
that as far as I could tell. I think something more generic is fine here.
We don't really care about the details.

xpc1000:
I finally put "DVB-TV" / "IPTV" / "Radio".

xpc1000:
Fixed error on x64 version and removed some warnings.

Underground78:
@xpc1000: It seems I forgot to post a message. I tried your test build some days and I had some issues with IPTV. However I'm not sure I understand right how I'm supposed to use it. My ISP provides rstp streams and an m3u playlist to group them all. How am I supposed to use that exactly?

xpc1000:
@Underground78: using IPTV is quite similar to using the current DVB-T UI. The stream is opened using the standard filter graph (the result should be similar to opening a stream from Fle->open file). There are some added functionality for multicasting and channel discovery but I believe it cannot be used with your provider. 

The main difference is  the way of adding new channels. If you go to options->Digital TV-> IPTV Channels you will find 4 methods for adding channels:

1. Import list: imports a m3u playlist that can be saved in the channel list
2. Manually: you can add a URL and the channel name manually
3. Range of IP's: It scans a range of IP addresses. Currently only for rtp, but we could add a dropdown with other possible protocols (udp, rstp, etc.). It manages IGMP membership automatically if the IP address is a valid multicast address.
4. Service discovery: automatically finds the list of channels from the Service Provider IP address when available. In this case, after suscribing to the multicast IP address, the application receives XML packets with the list of channels from a multicast address. It is currently restricted to rtp, but we could add an option to select between rtp and udp.

In your case, 1 and 2 would be the valid options. I could only test rtp so your feedback using rstp is important. I had to increase the parameter "Stream analisys duration" to 1400 in Lav Splitter configuration. Make sure your firewall is not blocking the communication.

Please let me know you feedback.

Underground78:
In fact I was able to get it working at least a bit but there are weird
things. The two most ennoying so far:
 - As far as I can tell many channels are missing after saving.
 - There is an UI glitch when double clicking on a channel, the navigator
bar disappears.

xpc1000:
The switching channel behavior is the same as DVB when rebuilding the filter graph. When switching channels, evrything is closed (including the panel) and then reopened. This is usually quite fast, but I believe in your case is quite slow for some reason. Could you confirm? (BTW double clicking is not necessary. One click is enough).

Regarding the channels while importing, could you send me a sample of the m3u file so I can test?

xpc1000:
@Underground78 : I added a new commit that changes the way recreate graph is addressed for both DVB and IPTV. Now the panel is not closed and reopened anymore.

You also mentioned a problem when importing channels from a m3u file. We still have the number of channels limited to 600. Is your list bigger than that? Could you share your list so I could test it?

Underground78:
The list is bigger than 600 but it wasn't the problem. Here is a sample from this playlist:
```
#EXTM3U
#EXTINF:0,1 - TF1 (TNT)
rtsp://192.168.0.9/fbxdvb/stream?tsid=6&nid=8442&sid=1537&frontend=1
#EXTINF:0,1 - TF1 HD (TNT)
rtsp://192.168.0.9/fbxdvb/stream?tsid=5&nid=8442&sid=1281&frontend=1
#EXTINF:0,2 - France 2 (bas débit)
rtsp://mafreebox.freebox.fr/fbxtv_pub/stream?namespace=1&service=201&flavour=ld
#EXTINF:0,2 - France 2 (auto)
rtsp://mafreebox.freebox.fr/fbxtv_pub/stream?namespace=1&service=201
#EXTINF:0,2 - France 2 (TNT)
rtsp://192.168.0.9/fbxdvb/stream?tsid=1&nid=8442&sid=257&frontend=1
#EXTINF:0,2 - France 2 (HD)
rtsp://mafreebox.freebox.fr/fbxtv_pub/stream?namespace=1&service=201&flavour=hd
#EXTINF:0,2 - France 2 (standard)
rtsp://mafreebox.freebox.fr/fbxtv_pub/stream?namespace=1&service=201&flavour=sd
#EXTINF:0,2 - France 2 HD (TNT)
rtsp://192.168.0.9/fbxdvb/stream?tsid=5&nid=8442&sid=1282&frontend=1
```

kasper93:
About the limit, we might refactor that to support more or unlimited list if there is a need for that. We probably should at some point if we are going to expand support for different media.

xpc1000:
@Underground78 : I fixed the parser. Could you please check again? Thanks

xpc1000:
@Underground78 , could u take a look? I fixed the parser and changed the way the graph is recreated when switching channels for both IPTV and DVB.

I also rebased to the latest version.

xpc1000:
Builds for those who want to test:
(32 bits) http://www.mediafire.com/download/sa96zx21scelhoz/mpc-hc.7z
(64 bits) http://www.mediafire.com/download/avd3d74ws0j066u/mpc-hc64.7z

Underground78:
If I press "IPTV Channels" I get an empty window.

xpc1000:
Fixed. The problem was when accessing for the first time only.

I updated the links:
(32 bits) http://www.mediafire.com/download/4cya2mtcmyrrp11/mpc-hc.7z
(64 bits) http://www.mediafire.com/download/0gy91mvnlz12ubn/mpc-hc64.7z

Underground78:
Ok I can confirm this problem is fixed. Now I have issue when switching
channels, they sometimes fail to open. It seems rather random. Also if I
want to retry opening this channel pressing play, it displays "playing" but
that's all.

xpc1000:
I assume you are talking about IPTV. Did you try increasing the parameter "Stream analisys duration" to at least 1400msecs in Lav Splitter configuration?

Underground78:
Yes and it doesn't seem to help. The issue might be related with the
switching order but I'm really not sure.

Also I think there are some race conditions in some cases because I have
seen weird things like playback restarting after being closed.

xpc1000:
I did some testing and I could reproduce your issue (only once). Unfortunately the problem seems to be very random so it is difficult to test. I did some changes in the last commit related to the process synchronization that probably solves the problem.

Underground78:
Which problem exactly? The race condition I assume. What about the other
one, do you reproduce it?

xpc1000:
Yes, the race condition.

I could reproduce the other one. The problem happens when trying to stop the FG before destroying the graph. More precisely when trying to enter a critical section in LavVideo.ax!CTransformFilter::Stop():

```C++
// Implements the CTransformFilter class
CTransformFilter::Stop()
{
    AutoLock lck1(&m_csFilter);
    if (m_State == State_Stopped) {
        return NOERROR;
    }
  
    // Succeed the Stop if we are not completely connected

    ASSERT(m_pInput == NULL || m_pOutput != NULL);
    if (m_pInput == NULL || m_pInput->IsConnected() == FALSE ||
        m_pOutput->IsConnected() == FALSE) {
                m_State = State_Stopped;
                m_bEOSDelivered = FALSE;
                return NOERROR;
    }

    ASSERT(m_pInput);
    ASSERT(m_pOutput);

    // decommit the input pin before locking or we can deadlock
    m_pInput->Inactive();

    // synchronize with Receive calls

    CAutoLock lck2(&m_csReceive);
...
```

Stack:

<Code>
 LAVVideo.ax!CCritSec::Lock() Line 499	C++
 LAVVideo.ax!CAutoLock::CAutoLock(CCritSec * plock) Line 98	C++
LAVVideo.ax!CTransformFilter::Stop() Line 471	C++
 [External Code]	
 mpc-hc64.exe!CMainFrame::OnPlayStop() Line 7277	C++
 mpc-hc64.exe!CMainFrame::ReCreateGraph() Line 14881	C++
 mpc-hc64.exe!CMainFrame::SetChannel(int nChannel) Line 15062	C++
 mpc-hc64.exe!CMainFrame::OnNavigateJumpTo(unsigned int nID) Line 8763	C++
 mpc-hc64.exe!CPlayerNavigationDialog::OnChangeChannel() Line 159	C++
</code>

Underground78:
Are you saying that it deadlocks? I'm not sure we are speaking of the same
issue.

xpc1000:
Yes. I get a black screen and there is still the message "playing". The player is not accepting any message.

xpc1000:
I get this after trying chanel swiching 10-15 times.

Underground78:
I think I'm also getting that issue in over circumstances but I wasn't
talking about this error here. Rather the case where opening the stream
fails with an error message in the status bar.

xpc1000:
OK now I see. 

When the open operation fails, it shows a message saying "Error when trying to open the stream" and some seconds later the timer overrides this message by showing the current status "Stopped". Switching to another channel is OK here, but I have to admit that I didn't review what happens when you press "play". I'll take a look to this.

The more frecuent cause for errors when opening the stream was due to the parameter  "Stream analisys duration". You might try to increse it to 2000. Does it happens the same when opening the same stream through "File->Open File..."?

Underground78:
I tried yesterday with higher value for "Stream analysis duration" but
without any luck. I cannot test now but tomorrow I will try opening the
link directly.

xpc1000:
When watching DVB, if you stop and then try to press "Play" has the same results. The solution will have to address both.

xpc1000:
I added a commit for solving the issue when pressing "play" when the player is stopped for both DVB and IPTV. It seems that the internal filters are not accepting the sequence of stopping and then playing a live stream. The solution I implemented was to simply avoid using the play button for digital TV. In case someone needs to restart playing, we can always switch to another (or now even the current) channel.

xpc1000:
Summary of the pending detected issues:

1. Opening a stream fails sometimes with an error message in the status bar: not sure what is happening here. @Underground78 can you check if the same problem happens when opening the link directly through File->Open file? 

2. When switching channels, sometimes the operation stays with a black screen. It seems there is a deadlock when trying to stop the filter graph (see https://github.com/mpc-hc/mpc-hc/pull/145#issuecomment-112928155)

xpc1000:
@Underground78 : I couldn't reproduce your error. Could you check opening the stream directly through File->Open File to see if the problem is more related to the internal splitter? Thanks

@kasper93 the last three commits affect DVB as well. Could you take a look? 

Thanks

xpc1000:
Updated to the latest version.

Links to testing builds:
(32 bits) http://www.mediafire.com/download/18mll1cr1rjrnb1/mpc-hc.7z
(64 bits) http://www.mediafire.com/download/856vasfgcfk5px1/mpc-hc64.7z

vBm:
IPTV works lovely. Manually imported chans one by one. Only "issue" I have is that when you press "+" to add new channel, chan list will not autoscroll so you end up having 30+ chans and first several are shown.

Also would be nice to have channel name in window title instead of just "Live".

THANKS @xpc1000 <3

xpc1000:
Thanks to you @vBm for testing!

Regarding the issues, I'll take a look this week. Both seem easy to address.

BTW adding more than 30 channels manually would have been a long process. You might consider editing a m3u file and then importing it. It would make things easier in case you had to repeat the process.

Jellyfrog:
If you pause IPTV you can't press play again

xpc1000:
The pause button is currently not active for DVB nor IPTV, since we don't have timeshift implemented. I assume you were referring to stop button. In case you press stop, you can start playing again by re-selecting the same channel. 

I disabled the play button because for some reason the internal filters are not accepting  stopping and then playing a live stream again. This happens both in DVB and IPTV.

xpc1000:
I did some small changes:
- fixed the autoscroll issue when adding channels manually. I also fixed it for channel discovery and for DVB channel scan. Now it scrolls down so you see the end of the list during the scanning process.
- IPTV: update the name of the channel in the window title (as done in DVB )
- Enable the OSD message with the channel name when switching a channel

xpc1000:
Links to test builds:
(32 bits) http://www.mediafire.com/download/18mll1cr1rjrnb1/mpc-hc.7z
(64 bits) http://www.mediafire.com/download/856vasfgcfk5px1/mpc-hc64.7z

vBm:
Thanks for changes, works well xD

One glitch I've noticed today. When you press stop, channel doesn't stop playing even tho OSD is displayed.

> IPTV Channel management: additional options (scan range of IP's and service discovery for multicast) are in progress and will be included in a different PR.

Please add export as well on that list :)

xpc1000:
> When you press stop, channel doesn't stop playing even tho OSD is displayed

Fixed

> > IPTV Channel management: additional options (scan range of IP's and service discovery for multicast) are in progress and will be included in a different PR.
> Please add export as well on that list :)

I'll put it on the list, but with less priority ;). 

BTW, scan range of IP's and service discovery are already working. Depending on the type of IPTV you are using you could try one of them.

vBm:
Is there a way to open EPG by any chance ? (or do you even plan to add support for it)

xpc1000:
> Is there a way to open EPG by any chance ? (or do you even plan to add support for it)

I planned  to start with a similar support to what we currently have for DVB for "now and next" and then make them grow together. Nevertheless I didn't see too much interest in merging this pull request after one year and a half, so for the moment development is almost frozen. 

This is the version that I'm using myself since more than one year so I try to keep it updated.

vBm:
Nice. Would love to see such feature in future :)

When it comes to review and merging, let's give a nudge to the team xD 
@XhmikosR @kasper93 @alexmarsev @Underground78

vBm:
@xpc1000 Any chance to get this rebased ? (I've failed to do it on my own and would really like new build xD)

xpc1000:
OK. I'll take a look...

xpc1000:
Rebased to the latest version.

Link to test build (64bits):
[(http://www.mediafire.com/file/856vasfgcfk5px1/mpc-hc64.7z)]

vBm:
Thanks @xpc1000 :)

xpc1000:
Rebased to the latest version.

lnxus:
Hello, I'm using the latest version posted here.  Everytime I restart MPC HC, I lose my saved IPTV streams (actually, one remains, the first one I added).  Any ideas?

xpc1000:
Hi, I'm quite busy now, but I'll take a look in a couple of weeks. In the mean time, you can try to set "Store settings in .ini file" in Options->Player. This is the way I use it and it works perfectly for me.

xpc1000:
I rebased and did some corrections.
@lnxus: could you test?

vBm:
Just did the testing and I don't lose the channel list upon restart of MPC-HC however current list of already saved channels isn't populated in the `IPTV channels management`.
That didn't work like ever so it's not a regression :)
Other than that it works quite well, so thanks @xpc1000 👍

vBm:
@xpc1000 could you possibly add logging info so we know when, why and which stream failed. Tapping into 762f0f2964b1a51b1d0f5f1cc3da020e21a4db5e ?

Eagle3386:
@xpc1000 Are you willing to adjust your awesome work like @clsid2 stated over in his fork https://github.com/clsid2/mpc-hc/issues/32?
If so, I'd happily assist in testing in case you need a tester as I happen to be an IPTV user.

PR #149, opened 2015-08-04
Supported Windows versions in README.md

ehea617:
Improves consistency of the two supported Windows version lists. Adds Windows 10 to the list.


XhmikosR:
Thanks for the PR but I'm gonna close it for now.

1. It's incomplete since we have docs/Readme.txt too
2. Has false info since we have some issues on Windows 10

We'll update the docs when everything is fine.

PR #150, opened 2015-08-11
Set the Video Window for EVR before connecting input pins

Nevcairiel:
This allows the video decoder to init DXVA2 decoding on the appropriate screen,
instead of having to re-init after starting the graph.

It also ensures the video decoding features of the adapter the window is currently on
are exposed, and not those of the default screen.

### DIFF
--- a/src/mpc-hc/FGFilter.cpp
+++ b/src/mpc-hc/FGFilter.cpp
@@ -479,6 +479,13 @@ HRESULT CFGFilterVideoRenderer::Create(IBaseFilter** ppBF, CInterfaceList<IUnkno
         if (m_clsid == CLSID_EnhancedVideoRenderer) {
             CComQIPtr<IEVRFilterConfig> pConfig = pBF;
             pConfig->SetNumberOfStreams(3);
+
+            if (CComQIPtr<IMFGetService> pMFGS = pBF) {
+              CComPtr<IMFVideoDisplayControl> pMFVDC;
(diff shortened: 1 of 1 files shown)


alexmarsev:
Makes sense as far as I can tell.

@Underground78, @kasper93 Your opinion?

kasper93:
I get that we already too late here https://github.com/mpc-hc/mpc-hc/blob/develop/src/mpc-hc/MainFrm.cpp#L11891 ? In this case this call should be removed and only new one should be left.

Nevcairiel:
It needs to be set before the pins are connected, so thats probably too late.
However, setting it again may not be a bad idea, as this function doesn't guarantee its set for every video renderer... there could be another custom one or something.

Underground78:
> @Underground78, @kasper93 Your opinion?

We already do this for madVR so I guess it's fine.

kasper93:
Could we make this fix also for EVR-CP and other renderers who implements this interface?

Underground78:
I don't think we support this.

kasper93:
Sure we do support it. The question is it needed.

Underground78:
I meant I'm not so sure we support setting the window handle so early.

Nevcairiel:
EVR-CP and Sync don't need this, as they get passed in the window handle when they are created and are automatically created on the right screen.

kasper93:
Ok.

PR #151, opened 2015-08-11
[RFC] Remove P010/P016 block from IPinHook

Nevcairiel:
P010 works in DXVA2-Native mode with EVR on some systems,
permanently blocking it would cripple that functionality.

Instead, LAV will block P010 memory buffers when connecting to EVR.

As an alternative, EVR could check if it was switched into DXVA2-Native mode before denying connection entirely.

Discuss!

### DIFF
--- a/src/filters/renderer/VideoRenderers/IPinHook.cpp
+++ b/src/filters/renderer/VideoRenderers/IPinHook.cpp
@@ -47,7 +47,6 @@ GUID g_guidDXVADecoder = GUID_NULL;
 int  g_nDXVAVersion = 0;
 
 IPinCVtbl* g_pPinCVtbl = nullptr;
-IPinCVtbl* g_pPinCVtblVideoDriverWorkAround = nullptr;
 IMemInputPinCVtbl* g_pMemInputPinCVtbl = nullptr;
 
 struct D3DFORMAT_TYPE {
@@ -149,19 +148,6 @@ static HRESULT STDMETHODCALLTYPE NewSegmentMine(IPinC* This, /* [in] */ REFERENC
     return NewSegmentOrg(This, tStart, tStop, dRate);
 }
 
-static HRESULT(STDMETHODCALLTYPE* ReceiveConnectionOrg)(IPinC* This, /* [in] */ IPinC* pConnector, /* [in] */ const AM_MEDIA_TYPE* pmt) = nullptr;
-
-static HRESULT STDMETHODCALLTYPE ReceiveConnectionMine(IPinC* This, /* [in] */ IPinC* pConnector, /* [in] */ const AM_MEDIA_TYPE* pmt)
-{
-    // Force the renderer to always reject the P010 and P016 pixel formats
-    if (pmt && (pmt->subtype == MEDIASUBTYPE_P010 || pmt->subtype == MEDIASUBTYPE_P016)) {
-        return VFW_E_TYPE_NOT_ACCEPTED;
-    } else {
-        return ReceiveConnectionOrg(This, pConnector, pmt);
-    }
-}
-
-
 static HRESULT(STDMETHODCALLTYPE* ReceiveOrg)(IMemInputPinC* This, IMediaSample* pSample) = nullptr;
 
 static HRESULT STDMETHODCALLTYPE ReceiveMineI(IMemInputPinC* This, IMediaSample* pSample)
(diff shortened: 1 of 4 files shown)


alexmarsev:
I guess it's about 10bit hevc.
Hmm, tricky.
> As an alternative, EVR could check if it was switched into DXVA2-Native mode before denying connection entirely.

I don't see how it can be done through IPin hooks, only through graph builder and I'd rather not touch it.

So unless I'm missing something, relying on LAVFilters for sanity checks is the most reasonable way we can go in my opinion.

kasper93:
> Instead, LAV will block P010 memory buffers when connecting to EVR.

I think we as MPC-HC need to keep this check for different decoders. (If there are any that support DXVA 10bit...)

> As an alternative, EVR could check if it was switched into DXVA2-Native mode before denying connection entirely.

I feel like this is better solution.

Nevcairiel:
As if there are any other decoders that use P010. ;)

kasper93:
> As if there are any other decoders that use P010. ;)

Yeah, right. I'm fine with removing this. But I don't want to reintroduce issues for people. @Underground78 should know better if it's safe.

> I don't see how it can be done through IPin hooks, only through graph builder and I'd rather not touch it.

One can dig into `AM_MEDIA_TYPE` and check `biCompression` if we are DXVA or not.

Nevcairiel:
> One can dig into AM_MEDIA_TYPE and check biCompression if we are DXVA or not.

No, that won't work. That media type is only refreshed afterwards to say DXVA, during connection its still NV12/P010.

What would need to be done is to hook IDirectXVideoMemoryConfiguration and intercept calls from the decoder that set the surface type to DXVA2_SurfaceType_DecoderRenderTarget .. but you would also need to reset this whenever the connection fails and whatnot, so its probably quite a load of house keeping - more than I would be comfortable to do in these hooks.

Nevcairiel:
As an alternative, you get a reference to the pin in the hook method anyway, so for all I care it could check if its LAV through the pin and allow it then.

Underground78:
> @Underground78 should know better if it's safe.

I don't really know. I am not aware of any decoders support P010/P016 but you never know...

> As an alternative, you get a reference to the pin in the hook method anyway, so for all I care it could check if its LAV through the pin and allow it then.

We could do that if we want to be sure. Not that a pin hook is particularly nice but unless the problem is fixed in the EVR (can it?), either we keep it or we bet on LAV being the only decoder that could cause issues.

Nevcairiel:
Microsoft seems unlikely to fix the problem in EVR, and with P010 being used for 10-bit DXVA, the GPU vendors better not turn it off again, or it'll break that. ;)

alexmarsev:
Quite frankly, I'd rather merge this PR as it is and don't try to fix EVR bugs in a roundabout way. That shouldn't affect generic playback scenarios.

@Underground78, @kasper93 Any objections?

Nevcairiel:
Note that you should update LAV before merging this, or you won't have the new block in yet.

kasper93:
I never liked this hook. But it was need. Anyway I'm fine with removing it. We can restore it if someone complains or think of better solution. I'm mostly woried about people using old LAV.

alexmarsev:
> Note that you should update LAV before merging this, or you won't have the new block in yet.

Ok.

> I'm mostly worried about people using old LAV.

And animu watchers with xy-vsfitler. Still we shouldn't assume responsibility that was never ours in the first place.

Underground78:
I agree with dumping the hook too but we need to be prepared for some
possible ranting I guess. Also I have LAV update ready but I think we
should wait for the standalone version of LAV to be released before
actually removing the hook. We need to be able to tell users to update
their standalone LAV if needed.

alexmarsev:
Did anyone investigate which gpu vendor caused the issue and reported it to them?

Nevcairiel:
Its a Microsoft bug, not GPU vendors.

alexmarsev:
Is it OS-specific? Because when I connect LAV Video Decoder (fairly old) to EVR in Graph Studio, it selects NV12 and not P010 for 10bit h264.

Nevcairiel:
Its not broken with every GPU or driver version, the GPU needs to accept P010 input to the video processor, which it does when it can do 10-bit decode.

alexmarsev:
So, only gtx960/skylake?

Nevcairiel:
Older Intel GPUs definitely have it enabled as well (Haswell and up at least). Dunno which NVIDIAs specifically, maybe only those with a 10-bit decoder, maybe more.

alexmarsev:
@Nevcairiel Does gtx960 supports 10bit hevc dxva decoding? I should have it around...

If it's indeed Microsoft bug, then we may have to use the hook after all :(
As it's unlikely that they are going to fix anything DirectShow-related.

Nevcairiel:
Yes, the 960 supports 10-bit DXVA. In Native mode as well if you apply my two PRs.

alexmarsev:
@Nevcairiel Hmm, NikosD person in your thread said that according to Intel the bug in EVR has been fixed in Win10, did anyone test that?

Nevcairiel:
It is not fixed.

Underground78:
@AlexMarsev: At some point the bug got fixed in some drivers but it was
reintroduced not long after, probably because manufacturers had a need for
P010. It makes the problem a bit difficult to follow but as far as I can
tell MS never did a thing about it.

alexmarsev:
@Underground78 As I understand it, the bug was never truly fixed, only worked around in the similar way as we do.

alexmarsev:
https://github.com/alexmarsev/mpc-hc/commit/c6426455dab94349235b635fcec890680df899a7 is probably a safe bet, as much as it pains me to have the hook around :)

Underground78:
Yeah, that was the other solution I had in mind if we decided not to remove
the hook.

kasper93:
At the beginning it was a bug in a driver, because no one had real 10-bit support and drivers claimed that they have. Anyway it is different story now.

I prefer Alex's solution. While it is true that the bug is not ours it will keep biting us. I still don't like the hook, but it is safer this way.

alexmarsev:
So I suppose https://github.com/alexmarsev/mpc-hc/commit/b25466c6c837f5230d048ff6a0b625423357d782 is the way we go.

Do we merge it now or wait for LAV stable?

@Nevcairiel Do you prefer to merge it now so you can receive some feedback from the community before releasing new LAV?

Nevcairiel:
I don't really care, its your call. Before the LAV stable you might have more users that accidentally run into the problem again.

Underground78:
I would prefer that we wait for LAV stable to limit the risks if that's ok
with everybody.

kasper93:
I don't think it is necessary to wait for stable release. People using nightly MPC-HC can very well use nightly LAV. Of course as long as LAV Stable release will come before MPC-HC's one

Nevcairiel:
A stable was released now, it would be great if this gets integrated soon, so 10-bit HW decoding actually works with MPC-HC.

Underground78:
I've updated my dev repo with all needed commits I think.

PR #153, opened 2015-11-09
Internal filters: Split and enable support for wmv3 decoding.

kongfl888:
Internal LAV Video Decoder also uses DMO, but it also supports MFT and hardware decoding.
It's useful for WMV-HD.


Underground78:
@Nevcairiel: Any opinion on whether we should enable WMV3 splitting and decoding with LAV Filters by default?

kongfl888:
Splitting wmv3 from wmv1/2/3 can allow users to have more choices.

kongfl888:
What do you think?

Nevcairiel:
@Underground78 There is no reason not to let LAV decode WMV3, it uses the MS MFT and as such is fully compatible, and even supports HWAccel on some hardware otherwise.

Underground78:
Ok, so your advice would be to let WMV1/2 decoding disabled but to enabled WMV3 decoding? What about splitting?

Nevcairiel:
There is no reason to not enable WMV1/2 as well. I'm not aware of any problems with it. Splitting the options may make sense, I suppose.

If you want to enable asf/wmv demuxing is up to you. There is pros and cons for both sides. Maybe leave it off and enable it next release, there is more fixes coming in future LAV versions.

Underground78:
I don't think I will make the changes now anyway, we will release a stable
too soon.

kasper93:
I'm not sure splitting those options make sense. We can just enable all of them.

Though it is either both splitting and decoding or nothing. I remember there was problems when mixing LAV Splitter and MS Decoder or other way around.

Underground78:
At least it used to be this way and I'm not completely sure if the splitter can be trusted. :/

PR #154, opened 2015-11-16
Add support for Chromecast devices.

theansible:
Not sure of this would have to be a plugin or a feature built into the program?


alexmarsev:
This is not a place for such discussion.

But a quick summary:
Chromecast unit is basically a web browser with some shortcuts and minor extra features. It is _extremely_ complex task to turn mpc-hc into web streaming service, what Chromecast support would require.

PR #155, opened 2015-12-02
VSFilter timestamp longlong

slydiman:
I have tested with these 2 commits and it works fine.

### DIFF
--- a/src/Subtitles/RTS.cpp
+++ b/src/Subtitles/RTS.cpp
@@ -2854,7 +2854,7 @@ STDMETHODIMP CRenderedTextSubtitle::NonDelegatingQueryInterface(REFIID riid, voi
 STDMETHODIMP_(POSITION) CRenderedTextSubtitle::GetStartPosition(REFERENCE_TIME rt, double fps)
 {
     int iSegment = -1;
-    SearchSubs((rt / 10000), fps, &iSegment, nullptr);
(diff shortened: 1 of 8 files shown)


Underground78:
I don't really see the point of this. int is enough to have more than 500h with a ms precision.

slydiman:
We have discussed it yesterday in IRC. I'm trying to use VSFilter in the StreamBufferEngine with absolute timestamps. This SBE uses the filetime format for timestamps (same 100ns units as reference time). So I got really huge numbers which are totally corrupted inside VSFilter. Everything works fine with SBE after these fixes.

kasper93:
> I don't really see the point of this. int is enough to have more than 500h with a ms precision.

Eh, I said exactly that on IRC yesterday. :) But if it is needed I'm fine with changing that.

@slydiman: Please use `REFERENCE_TIME` type instead of `LONGLONG`. Also if you are changing those types you need to change everything related. Also I wouldn't convert times to milliseconds anymore if we use bigger data type. Rater use RT everywhere.

slydiman:
I used LONGLONG instead of REFERENCE_TIME because of milliseconds. 
Ok, I will update it to RT and no milliseconds anymore.

slydiman:
Complete

slydiman:
Just wondering... Has anyone had a chance to review it? Thanks.

PR #158, opened 2015-12-27
Divide by 0 workaround (SeparableFilterX_SSE2)

hawky358:
The divisor can become 0. When this happens it causes
SeparableFilterX_SSE2 to divide by zero resulting in a crash.
( https://drdump.com/DumpGroup.aspx?DumpGroupID=400433 )

This workaround ensures SeparableFilterX_SSE2 does not receive a 0
divisor thus preventing a crash.

### DIFF
--- a/src/Subtitles/SeparableFilter.h
+++ b/src/Subtitles/SeparableFilter.h
@@ -361,6 +361,7 @@ struct GaussianKernel {
         for (int x = width / 2 - 1; x >= 0; x--) {
             short val = (short)(NormalDist(sigma, width / 2 - x) * 255 + 0.5);
             divisor += val * 2;
+			if (divisor == 0) divisor = 1;
             kernel[x] = val;
             kernel[width - x - 1] = val;
         }


kasper93:
Can you reproduce this issue? Do you have a sample subtitle file? While your commit would workaround division by 0 we need to diagnose why this actually happens at this point. In crash dumps from this problem I see `m_style.fGaussianBlur` to be `7.4775600000000001e-308`. I would like to debug why this happens and fix in the right place.

Underground78:
I cannot really check now but in theory I don't think the divisor can
actually be 0.

hawky358:
Hello,

"Can you reproduce this issue? Do you have a sample subtitle file?"

I've managed to isolate the issue to a specific line within a subtitle. This is just one, I've had this issue several times before. I can probably try to find more if needed. 
[Isolate Issue.zip](https://github.com/mpc-hc/mpc-hc/files/73354/Isolate.Issue.zip)

I've moved the subtitle to around 2 minutes. Too early and the video crashes immediately.
If you seek to around there it will crash or just let it run until it gets close then it will crash.


"While your commit would workaround division by 0 we need to diagnose why this actually happens at this point" 

That would be ideal. I'm not a coder/programmer and have basically 0 experience with C++ so I don't know where to start, the crashes just annoyed me so I made it work.

"I cannot really check now but in theory I don't think the divisor can actually be 0."
That's what it ends up as when debugging. 
Since m_style.fGaussianBlur goes to 7.4775600000000001e-308 (from what kasper93 says) it and divisor is an integer, it probably goes to 0 after the normal distribution function. I'll leave that to the experts to figure out.


Let me know if you need anything else.

kasper93:
> I've managed to isolate the issue to a specific line within a subtitle. This is just one, I've had this issue several times before. I can probably try to find more if needed.
Isolate Issue.zip

Thanks I will take a look later.

---------------

`if (fGaussianBlur > 0) {` is wrong. It will be true even for very small values. And later `divisor` will be 0 if we calculate for floating point values close to zero. I will make a patch for this issue.

kasper93:
This is limitation of our current fixed point implementation. Blur values higher than 203.45 doesn't work. I've created a ticket about this issue https://trac.mpc-hc.org/ticket/5729 

I'm closing this pull request, because fix provided is not sufficient and this is not the right place for bug reports.

PR #161, opened 2016-02-29
Shader Cache

on-situ-repos:
Since some of our shaders take up to one minute to build, I added a shader cache for the PixelShaderCompiler :

- compiled shaders are stored locally, in USER_APP_DATA (ex: C:\Users\ ***** \AppData\Roaming\\MPC-HC\ShaderCache ) or folder set by ini file
- hash of the cached shader is computed with shader profile and shader code
- cached shader is removed if it has not been used for 30 days, or the option disabled
- added a checkbox in the Output configuration, enabled for VMR/EVR (default is not active)
- handle folder deletion in uninstall process


src/filters/renderer/VideoRenderers/PixelShaderCache.cpp
```diff
@@ -0,0 +1,299 @@
+/*
+* (C) 2006-2016 see Authors.txt
```
XhmikosR: Since these are new files, just use the current year.

XhmikosR:
Please check if we need to update the installer script to remove the new folder on uninstall.

on-situ-repos:
Right, I added it in the InnoSetup config.

src/filters/renderer/VideoRenderers/VideoRenderers.vcxproj
```diff
@@ -64,6 +64,7 @@
     <ClCompile Include="madVRAllocatorPresenter.cpp" />
     <ClCompile Include="OuterEVR.cpp" />
     <ClCompile Include="OuterVMR.cpp" />
+	<ClCompile Include="PixelShaderCache.cpp" />
```
XhmikosR: Tabs are bad.

XhmikosR:
Why did you choose localappdata? We already have the userappdata folder.

Also, I believe the folder name and new settings should be in SettingsDefines.h.

src/filters/renderer/VideoRenderers/PixelShaderCache.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+        fclose(fp);
+
+        return ret;
+    }
+    return false;
+}
+
+void CPixelShaderCache::DeleteCache(
+    unsigned int hash)
```
XhmikosR: Maybe keep it one line here since it's not long.

on-situ-repos:
Ok, I'll replace it with DelTree then, and test. I'll push the updates tomorrow.
They are subtle differences between Local & Roaming folders ( http://superuser.com/questions/150012/what-is-the-difference-between-local-and-roaming-folders ), that's why I choose the Local one. Maybe it's not really required, as you wish.

XhmikosR:
Just don't use Del Tree in the root mpc-hc folder. Better remove the
subfolder and then the root if it's empty. The latter won't be necessary if
you just use the same folder we use already.

alexmarsev:
I guess conceptually it all makes sense (although I'm a bit curious about what shaders take up to one minute to compile).

My only point of concern is not very collision resilient 32bit-long hash. Probably better to make it _at least_ 64bit-long.

on-situ-repos:
We use mpc for multi-projectors installations, sometimes coupled with projection on 3d objects (heightmaps...), so yes the resulting shaders may be heavy. You are right about the 64 bits hash.

src/mpc-hc/AppSettings.cpp
```diff
@@ -1922,6 +1924,13 @@ void CAppSettings::UpdateRenderersData(bool fSave)
             delete [] dPtr;
         }
 
+        ars.bCacheShaders = !!pApp->GetProfileInt(IDS_R_SETTINGS, IDS_RS_CACHESHADERS, DefaultSettings.bCacheShaders);
+        if (AfxGetMyApp()->GetAppDataPath(ars.sShaderCachePath)) {
```
kasper93: Please use `GetAppSavePath()` if .ini file is used cache should be stored in the same folder as binary.

src/mpc-hc/AppSettings.cpp
```diff
@@ -1922,6 +1924,13 @@ void CAppSettings::UpdateRenderersData(bool fSave)
             delete [] dPtr;
         }
 
+        ars.bCacheShaders = !!pApp->GetProfileInt(IDS_R_SETTINGS, IDS_RS_CACHESHADERS, DefaultSettings.bCacheShaders);
+        if (AfxGetMyApp()->GetAppDataPath(ars.sShaderCachePath)) {
+            CPath path;
+            path.Combine(ars.sShaderCachePath, IDS_R_SHADER_CACHE);
+            ars.sShaderCachePath = (LPCTSTR)path;
```
kasper93: Replace those three lines with our helper `PathUtils::CombinePaths`

src/mpc-hc/mpc-hc.rc
```diff
@@ -2532,6 +2533,12 @@ END
 
 STRINGTABLE
 BEGIN
+	IDC_CACHESHADERS
```
kasper93: indentation is wrong here

src/mpc-hc/mpc-hc.rc
```diff
@@ -803,6 +803,7 @@ BEGIN
     ICON            "",IDC_QTRND_SAVEIMAGE_SUPPORT,221,75,20,20
     GROUPBOX        "Subtitle Renderer",IDC_STATIC,5,121,285,30
     COMBOBOX        IDC_COMBO1,10,133,276,46,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+	CONTROL         "Cache compiled Shaders",IDC_CACHESHADERS,"Button",BS_AUTOCHECKBOX | BS_MULTILINE | WS_TABSTOP,10,232,171,10
```
kasper93: spaces, not tabs

src/filters/renderer/VideoRenderers/VideoRenderers.vcxproj
```diff
@@ -90,6 +91,7 @@
     <ClInclude Include="madVRAllocatorPresenter.h" />
     <ClInclude Include="OuterEVR.h" />
     <ClInclude Include="OuterVMR.h" />
+	<ClInclude Include="PixelShaderCache.h" />
```
kasper93: spaces, not tab.

src/filters/renderer/VideoRenderers/PixelShaderCache.cpp
```diff
@@ -0,0 +1,270 @@
...
+    LPCSTR pSourceData,
+    SIZE_T SourceDataSize,
+    LPCSTR pSourceName) const
+{
+    return Hash(pProfile) + Hash(pSourceData) + (uint64_t)SourceDataSize + Hash(pSourceName);
+}
+
+uint64_t CPixelShaderCache::Hash(LPCSTR pStr) const
+{
+    uint64_t hash = 3074457345618258791ul;
```
kasper93: Postfix should be either `ull` or `Ui64`.

src/filters/renderer/VideoRenderers/PixelShaderCache.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+
+bool CPixelShaderCache::LoadCache(const CString& FileName, const CString& FilePath)
+{
+    if (IsFileOutdated(FilePath)) {
+        return false;
+    }
+
+    uint64_t hash = _tcstoui64(FileName, NULL, 10);
+    if (FILE* fp = _tfsopen(FilePath, _T("rt"), _SH_SECURE)) {
```
kasper93: Have you considered using `CFile` class?

src/filters/renderer/VideoRenderers/PixelShaderCache.cpp
```diff
@@ -0,0 +1,270 @@
...
+        FILETIME file_time;
+        if (SystemTimeToFileTime(&current_time, &file_time)) {
+            SetFileTime(file_handle, NULL, NULL, &file_time);
+        }
+
+        CloseHandle(file_handle);
+    }
+}
+
+bool CPixelShaderCache::IsFileOutdated(const CString& FileName) const
```
kasper93: This whole method could be greatly simplified by using `CFile` class with `GetStatus` and comparing `CTime`.

src/filters/renderer/VideoRenderers/PixelShaderCache.cpp
```diff
@@ -0,0 +1,270 @@
...
+    SIZE_T SourceDataSize,
+    LPCSTR pSourceName,
+    IDirect3DPixelShader9** ppPixelShader)
+{
+    if (!IsEnabled()) {
+        return E_FAIL;
+    }
+
+    uint64_t hash = Hash(pProfile, pSourceData, SourceDataSize, pSourceName);
+    CMap<uint64_t, uint64_t, CAutoVectorPtr<DWORD>, CAutoVectorPtr<DWORD>>::CPair* pixelShader = m_PixelShaders.PLookup(hash);
```
kasper93: This is long name, could use `auto` instead :)

src/filters/renderer/VideoRenderers/PixelShaderCache.cpp
```diff
@@ -0,0 +1,270 @@
...
+    : m_pD3DDev(pD3DDev)
+{
+    if (IsEnabled()) {
+        LoadCache();
+    } else {
+        DestroyCache();
+    }
+}
+
+CPixelShaderCache::~CPixelShaderCache()
```
kasper93: I guess no need to define empty destructor. Default one will do the job.

kasper93:
> I guess conceptually it all makes sense (although I'm a bit curious about what shaders take up to one minute to compile).

This is indeed extreme example, but such cache will help in all cases I guess. Even if the gain will not be one minute I think it is good inclusion.

src/filters/renderer/VideoRenderers/PixelShaderCache.cpp
on-situ-repos: Ok, didnt know it (_tfsopen is used in Shaders.cpp). Thanks.

src/filters/renderer/VideoRenderers/PixelShaderCache.cpp
on-situ-repos: Ok, great!

alexmarsev:
@on-situ-repos 
What is the reason behind 10MB limit? Also please add .cso extension (official) to cache objects and I think we can merge this.

Underground78:
Also please squash the commits (at least some).

on-situ-repos:
ok, i removed the 10MB limit, the time limit do the job.

Underground78: I create a new branch and will push a single commit with the cache.

Underground78:
You can force push to this branch.

on-situ-repos:
Ok i squashed / rebased the changes & force push, thanks.

Underground78:
Closed via 32389c59924ad329a98570d41c52b6752f9c5fbe.

PR #162, opened 2016-03-04
Shader media times access

on-situ-repos:
Add media status access from shader, add current video time and length in the const registers.


src/filters/renderer/VideoRenderers/DX9RenderingEngine.cpp
```diff
@@ -390,20 +390,28 @@ HRESULT CDX9RenderingEngine::RenderVideoDrawPath(IDirect3DSurface9* pRenderTarge
...
+        const float clTickToSeconds = 1.0f / 10000000.0f;
+
 #if 1
         D3DSURFACE_DESC desc;
         m_pVideoTexture[m_nCurSurface]->GetLevelDesc(0, &desc);
 
         float fConstData[][4] = {
-            {(float)desc.Width, (float)desc.Height, (float)(counter++), (float)diff / CLOCKS_PER_SEC},
-            {1.0f / desc.Width, 1.0f / desc.Height, 0, 0},
+            { (float)desc.Width, (float)desc.Height, (float)(counter++), (float)diff / CLOCKS_PER_SEC },
```
XhmikosR: Remove these, use contrib/run_astyle.bat for formatting.

XhmikosR:
You do not need the merge commits; just rebase your branch.

src/filters/renderer/VideoRenderers/DX9RenderingEngine.cpp
on-situ-repos: I do run it before pushing, but it doesn't correct it. I use AStyle 5.5.8 unicode, is this correct ?

on-situ-repos:
Ok, I haven't used "rebase" for now, thanks for the tip.

src/filters/renderer/VideoRenderers/DX9RenderingEngine.cpp
XhmikosR: That version doesn't exist? Personally, I'm still in 2.05.1, no idea if there's any newer, but you shouldn't do manual corrections in style since it will be overwritten the next time.

on-situ-repos: Ok, sorry, it's 2.05.1.

XhmikosR: The issue might have been already fixed in their trunk branch, but you can try reporting it to astyle.

Underground78: If I understand correctly you are trying to give the shaders access to the
current frame time. I'm actually not sure that what you do is correct. It
can sort of work but it seems hacky to rely on the info gathered from
CMainFrame. There must be cleaner, safer, more accurate way to do that
directly from the renderer.

on-situ-repos: I'm not sur if it's a bug : --unpad-paren seems to be about parenthesis only, not brackets.

Yes I guess the method is intrusive, but we didn't see other way to retrieve the information.

alexmarsev: No, gathering stream position information on timer is way too imprecise. And calling IMediaSeeking from rendering loop would cause deadlock, if I remember DirectShow quirks correctly. But video renderer should have segment position for each to be presented frame, and that can be used.

@on-situ-repos 
Please clarify why do you need this in the first place, I think it's getting out of scope of what video renderer shaders were supposed to do.

Underground78: Yes, that's exactly what I meant.

on-situ-repos: I'll look at the segment position.
It's to have effects based on the timeline (custom fadeoff / transitions / ...).
We are working on adding an external .json to set custom shader attributes (textures & others), not sure if it should be pushed too, it'll be on github anyway.

alexmarsev: > We are working on adding an external .json to set custom shader attributes (textures & others), not sure if it should be pushed too, it'll be on github anyway.

If you can wrap it into an interface that could be used by external video renderers such as madVR, then by all means yes.

on-situ-repos:
Ok, last commit is better.
Here is a sample shader for testing : 

float4 consts0 :  register(c0);
float4 consts1 :  register(c1);

float4 main(float2 tex : TEXCOORD0) : COLOR
{	
	return float4(consts1.z/4.0, (consts1.w-consts1.z)/4.0, 0.0, 1.0);
}

src/SubPic/ISubPic.h
```diff
@@ -183,7 +183,8 @@ public IUnknown {
     STDMETHOD_(bool, Paint)(bool bAll) PURE;
 
     STDMETHOD_(void, SetTime)(REFERENCE_TIME rtNow) PURE;
-    STDMETHOD_(void, SetSubtitleDelay)(int delayMs) PURE;
+    STDMETHOD_(void, SetDuration)(REFERENCE_TIME rtDuration) PURE;
+	STDMETHOD_(void, SetSubtitleDelay)(int delayMs) PURE;
```
vBm: tabs -> spaces
Maybe run AStyle if you haven't already.

on-situ-repos: ISubPic.h is excluded by astyle.ini

PR #165, opened 2016-04-05
Readme & Authors: some corrections

blackarcher21:
That's all that I could find and made some correction

- On SourceForge, its name is XmlRpcC4Win, on GitHub it's XmlRpc4Win. Not sure which one is right though.
- Changed Transifex link to direct link.

That's all. regards. :)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -85,7 +85,7 @@ MPC-HC makes use of the following third-party code:
 | TreePropSheet     | -                         | http://www.codeproject.com/Articles/3709/ |
 | UnRAR             | freeware                  | http://www.rarlab.com/rar_add.htm |
 | VirtualDub        | GPLv2+                    | http://www.virtualdub.org/ |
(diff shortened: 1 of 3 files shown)


XhmikosR:
Thanks for this. I have picked your Transifex change for now.

I'm not sure which name to use either. :/

/CC @drtimcooper: is it XmlRpcC4Win or XmlRpc4Win?

drtimcooper:
@XhmikosR @blackarcher21 It's the same project.  I migrated from SourceForge to github recently.  So please call it XmlRpc4Win because the github version is the one I'll be maintaining.  (For the record, the extra 'C' meant "client" because it implements only the client side of XmlRpc. I forget if dropping the extra 'C' was intentional or accidental).

XhmikosR:
@drtimcooper: thanks for the reply!

@blackarcher21: closing this, and I'll push your patch manually in the next batch of patches.

PR #166, opened 2016-06-08
Add support for AssFilter DirectShow subtitle provider

Blitzker:
I don't know well the mpc-hc code base, but thoses changes add the "AssFilter" subtitle renderer in the Options -> Output -> Subtitle Renderer.


src/mpc-hc/resource.h
```diff
@@ -1569,9 +1569,10 @@
 #define IDS_PPAGEADVANCED_SCORE         57530
 #define IDS_PPAGE_FS_CLN_AUDIO_DELAY    57531
 #define IDS_PPAGEADVANCED_DEFAULTTOOLBARSIZE 57532
+#define IDS_SUBTITLE_RENDERER_ASS_FILTER 57533
 
 // Next default values for new objects
-//
```
kasper93: Please revert this trailing space. It is added by visual studio resource editor. So it needs to stay so it won't pop every time someone edits it.

Blitzker:
Sorry about that. I also added commit 01457be so AssFilter is the only active subtitle renderer when it is selected.

src/mpc-hc/resource.h
Blitzker: Sorry about that. Fixed

amayra:
i like to see AssFilterMod in mpc-hc

PR #167, opened 2016-10-21
Save Image (auto): if directory isn't available, Save Image instead

JustinHydroxide:
This addresses [#5862](https://trac.mpc-hc.org/ticket/5862). Please let me know if there's anything else I should do (e.g. edit the ticket). I'm new here but I'd like to contribute whatever I can.


kasper93:
@JustinHydroxide: Thanks!

PR #168, opened 2016-11-09
Add support for AssFilter DirectShow subtitle provider

Blitzker:
Instead of updating my old pull request, i decided to close it and make a new one instead. Its more polished and the changes are contained in 1 commit.

This adds a new item, AssFilter,  in Options->Playback->Output->Subtitle Renderer.

If AssFilter is selected and Options->Advanced->BlockVSFilter is True, the other subtitles renderers are blocked from loading.


amayra:
when i can use AssFilter without hack ?

kasper93:
Thank you!

PR #169, opened 2016-11-10
Skipping back in a file with no chapters seeks to start instead of skipping to previous file

JustinHydroxide:
If you tried to skip back in a file with no chapters, it started playing the previous file instead of seeking to the start of the current one. I wasn't sure if this was by design, but it bothered me personally since it felt inconsistent with files with chapters.


XhmikosR:
There's an option to play the next/previous file in folder. How exactly do your changes take that into consideration?

JustinHydroxide:
I don't believe the function that I edited is called in that case, if you're talking about the "Next File" and "Previous File" commands that can be configured with hotkeys.

PR #172, opened 2017-02-22
If no index with hwgpu is sent, leave the value of iLAVGPUDevice as is

precla:
Fixes #5881


XhmikosR:
Is this right? `iLAVGPUDevice` seems to be a DWORD.

precla:
Hmm, going to check that next week. At least it doesn't crash mpc-hc now, no matter what you add after '/hwgpu', even if you don't add anything.
Will respond as soon as i test it again.

precla:
@XhmikosR DWORd is just a unsinged long, so setting it to 0 (NULL) should be ok. I tested it, with sending one arg (/hwgpu) and multiple (/hwgpu) and nothing crashes. Can you give me more info why this should be a problem? Thanks

PR #174, opened 2017-03-09
Changed references of "Pan-n-Scan" to "Pan&Scan" for consistency.

cunnj:
Also fixed incorrect capitalization of "to".

### DIFF
--- a/src/mpc-hc/mpc-hc.rc
+++ b/src/mpc-hc/mpc-hc.rc
@@ -182,7 +182,7 @@ FONT 8, "MS Shell Dlg", 400, 0, 0x1
 BEGIN
     ICON            "",IDR_MAINFRAME,14,10,48,48
     LTEXT           "",IDC_STATIC1,60,8,200,8
-    CONTROL         "Copyright © 2002-2017 see Authors.txt",IDC_AUTHORS_LINK,
+    CONTROL         "Copyright Â© 2002-2017 see Authors.txt",IDC_AUTHORS_LINK,
                     "SysLink",WS_TABSTOP,60,20,200,8
     CONTROL         "",IDC_HOMEPAGE_LINK,"SysLink",WS_TABSTOP,60,32,200,8
     LTEXT           "This program is freeware and released under the GNU General Public License.",IDC_STATIC,7,46,246,18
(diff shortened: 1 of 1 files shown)


XhmikosR:
@kasper93: this needs to be merged manually if you agree.

Also, I don't get the degree change; the characters look identical to me.

johnc7995:
I didn't change the degree symbol, at least not voluntarily.

XhmikosR:
Can you make the changes again with a proper code editor?

kasper93:
Looks ok. I will merge it with next translation update. Of course with fixed encoding.

matthewmauger:
Looks fine, though the degree thing is weird.

@XhmikosR - why does this need a manual merge?

XhmikosR:
@matthewmauger: the resources need to be updated too :)

The scripts are here https://github.com/mpc-hc/mpc-hc/tree/develop/src/mpc-hc/mpcresources

Basically needs https://github.com/mpc-hc/mpc-hc/blob/develop/src/mpc-hc/mpcresources/UpdateIS.bat after this patch is merged so that the translations pick up the changes.

We use Transifex for translations BTW.

matthewmauger:
@XhmikosR: Ah, understood. Forgive my ignorance but should we approve, let the change get pushed, manually pull develop, fix the resources and push back the adjusted resource files?

I'll approve this change then but I don't think I currently have the rights to do anything more than that.

matthewmauger (approved):
Approved, looks fine and sounds like the resource update is a separate step that needs to be done after your changes are in.

XhmikosR:
Yeah but since the encoding changes aren't made you shouldn't approve it.

matthewmauger (changes requested):
Sure, waiting for translation changes as requested.

@XhmikosR, you'll have to explain to me exactly when those translation changes get merged in - are you expecting the original author to do it or someone with write access to develop?

XhmikosR:
I tried to manually merge this but then all translations for those strings will be reset so I'm not sure if it's worth the change. Alternatively we could do a search and replace manually.

kasper93:
Thanks!

blackarcher21:
for Pan&Scan,  I just recheck latest nighlty (1.7.13.60) it become acceleration key instead of showing the "&" character.

PR #175, opened 2017-03-24
ParseCommandLine: Add Command for Muting Audio

precla:
Fixes #5657
Add 'mute' switch to launch File with muted Audio.


src/mpc-hc/AppSettings.cpp
```diff
@@ -2105,6 +2105,8 @@ void CAppSettings::ParseCommandLine(CAtlList<CString>& cmdln)
                 nCLSwitches |= CLSW_ADD;
             } else if (sw == _T("randomize")) {
                 nCLSwitches |= CLSW_RANDOMIZE;
+            } else if (sw == _T("mute")){
```
kasper93: Why add here? Keep some sane order. Probably end will be good to make it roughly follow enum order.

src/mpc-hc/MainFrm.cpp
```diff
@@ -11189,6 +11189,10 @@ void CMainFrame::OpenSetupAudio()
 {
     m_pBA->put_Volume(m_wndToolBar.Volume);
 
+    if (AfxGetAppSettings().nCLSwitches & CLSW_MUTE) {
```
kasper93: Probably should be before put_Volume. Also `OpenSetupAudio()` is called when starting playback. I think it should set volume not only in this case. That say probably better place to set vol is when restoring setting. https://github.comhttps://github.com/mpc-hc/mpc-hc/blob/develop/src/mpc-hc/AppSettings.cpp#L760

(simply set fMute when parsing cmdline)

precla: That is in SaveSettings() , do you mean LoadSettings() ?
https://github.com/mpc-hc/mpc-hc/blob/develop/src/mpc-hc/AppSettings.cpp#L1336

src/mpc-hc/AppSettings.cpp
precla: Changed order

matthewmauger (changes requested):
Generally fine but this will now clash with recent changes as there is a new resource entry with the same ID as yours. Can you fix this, put yours in the proper location and resubmit?

precla:
Close this mumbo-jubmo.. Made a new clean one: https://github.com/mpc-hc/mpc-hc/pull/184

PR #176, opened 2017-03-25
ParseCommandLine: Add Command for setting Volume

precla:
Feature request trac: #5991
Add 'volume' switch to launch File with desired volume level.


src/mpc-hc/MainFrm.cpp
```diff
@@ -11189,6 +11189,10 @@ void CMainFrame::OpenSetupAudio()
 {
     m_pBA->put_Volume(m_wndToolBar.Volume);
 
+    if (AfxGetAppSettings().nCLSwitches & CLSW_VOLUME) {
```
kasper93: Why this is needed? You already set volume in setting...

src/mpc-hc/AppSettings.cpp
```diff
@@ -2105,6 +2105,12 @@ void CAppSettings::ParseCommandLine(CAtlList<CString>& cmdln)
                 nCLSwitches |= CLSW_ADD;
             } else if (sw == _T("randomize")) {
                 nCLSwitches |= CLSW_RANDOMIZE;
+            } else if (sw == _T("volume")) {
```
precla: Added

src/mpc-hc/MainFrm.cpp
precla: Without it, it doesn't set the Volume at all. Only on 2nd launch of mpc-hc. I don't know why and what I did wrong

matthewmauger: Possible race condition due to that line getting hit from the graph thread before the toolbar volume has initialised?

kasper93:
Thanks!

PR #177, opened 2017-03-26
ParseCommandLine: Add Command for View Preset

precla:
Feature request trac: https://trac.mpc-hc.org/ticket/3475
Add 'viewpreset' switch to launch MPC-HC with one of the 3 Presets
1 -> minimal
2 -> compact
3 -> normal


XhmikosR:
All this are nice, but they aren't added to the help dialog so it's basically hidden.

Not sure we want to bloat the dialog more, though.

precla:
Yeah, quite a bunch of command switches already there (cl dialog).
Adding them wouldn't be a problem, but do we want to add them there?
Maybe we should clean up the Feature requests on trac before we move on with such things.
Maybe see what others think of more cl switches?

kasper93:
>  Adding them wouldn't be a problem, but do we want to add them there?

Sure. There is literary no point in adding command line switches if we don't have them documented anywhere. Command line help is not perfect, but it can hold more options.

precla:
Added into CL Help dialog

matthewmauger (approved):
Looks good to me! Only issue is whitespace differences which looks like it's causing a conflict.

matthewmauger (changes requested):
Ok looks like this is not going to go in without a merge - can you pull develop, merge into your branch fixing the clash in src/mpc-hc/resource.h and resubmit?

Edit: Just noticed that your resource entries are also out of date. You're going to need to pull latest, update your string - fix the carriage returns or whatever is making the diff fail - and resubmit. At the moment we're going to have a clash due to duplicate IDs. This is also an issue with your other pull requests which create different entries but all of which use the same ID (57535)

precla:
I'll have a look into it in the next few days.

src/mpc-hc/resource.h
```diff
@@ -1540,6 +1540,7 @@
 #define IDS_CMD_MONITOROFF              57502
 #define IDS_CMD_PLAYNEXT                57503
 #define IDS_CMD_FULLSCREEN              57504
+<<<<<<< HEAD
```
XhmikosR: This is definitely wrong :P

precla:
@matthewmauger Everything done for this. Thanks to @XhmikosR 
Should be ok now. Can you check please?

matthewmauger (approved):
Looks good to me! Thanks for fixing that up. Approved.

kasper93:
Thanks!

PR #179, opened 2017-04-09
fix: textsub renders nothing in avisynth.

sorayuki:
when used as filter in avisynth, it renders nothing in video.

### DIFF
--- a/src/filters/transform/VSFilter/plugins.cpp
+++ b/src/filters/transform/VSFilter/plugins.cpp
@@ -86,7 +86,8 @@ namespace Plugin
 
             if (!m_pSubPicQueue) {
                 CComPtr<ISubPicAllocator> pAllocator = DEBUG_NEW CMemSubPicAllocator(dst.type, size);
-
+                pAllocator->SetCurVidRect(CRect(CPoint(), size));
+                
                 HRESULT hr = E_FAIL;
                 if (!(m_pSubPicQueue = DEBUG_NEW CSubPicQueueNoThread(SubPicQueueSettings(0, 0, false, 50, 100, false), pAllocator, &hr)) || FAILED(hr)) {
                     m_pSubPicQueue = nullptr;


git24hub:
The classic player can't play reverse? only forward...

sorayuki:
Maybe I didn't explain it clearly. Please let me throw light on the patch again.
MPC-HC publishes binaries in two ways; one is player executable file and the other is standalone filters.
In the package of standalone filters there is a dynamic linked library called "VSFilter.DLL"
It used to work as AviSynth plugin to overlay subtitles on video.
In the distribution of MPC-HC's standalone filters, it fails to render subtitles on video when used as AviSynth plugin.
I submitted this patch to fix the problem.

vBm:
@sorayuki please ignore @git24hub, It's just a spam. Pretty much the same text in other place.
@XhmikosR will delete that ASAP.

Thanks for the patch.

jack17529:
And the patch still did not got merged ?

XhmikosR:
No one has reviewed this and it's basically one person who is still active
(sort of).

sorayuki:
Hello, has the reviewer really seen this pull request?

XhmikosR:
Nope. There's no one around active, unfortunately. If any of the project follower's wants to contribute by reviewing the PRs or anything else, let me know.

JodiTheTigger:
I would check the pointer for null before dereferencing it, `CSubPicQueueNoThread` can handle the allocator being null, but your code cannot.

matthewmauger (changes requested):
Ok I think there's an underlying problem here.

Both CMemSubPicAllocator and it's base class CSubPicAllocatorImpl have the same data member defined in each class - m_curvidrect. That's obviously bad practice, though it is valid C++ as both are private.

If even both members having the same member is a good idea (which I seriously doubt) the proper fix would be to rename the one in the derived class and initialise it properly in the CMemSubPicAllocator constructor, just like CSubPicAllocatorImpl does.

More likely, the derived class simply should never have created it's own copy of the variable with the same name.

sorayuki, on the chance you haven't totally given up on this change could you please fix this in the constructor of CMemSubPicAllocator like this:

CMemSubPicAllocator::CMemSubPicAllocator(int type, SIZE maxsize)
    : CSubPicAllocatorImpl(maxsize, false)
    , m_type(type)
    , m_maxsize(maxsize)
{
	m_curvidrect = CRect(CPoint(0, 0), m_maxsize);
}

As this will fix this problem for all other client code that uses it, rather than having to call another function as a hack after instantiating it.

Thanks!

kasper93:
Would be better to just remove this duplicated value. 

``` diff
diff --git a/src/SubPic/MemSubPic.cpp b/src/SubPic/MemSubPic.cpp
index defab054d..5e0c9aace 100644
--- a/src/SubPic/MemSubPic.cpp
+++ b/src/SubPic/MemSubPic.cpp
@@ -725,9 +725,3 @@ STDMETHODIMP CMemSubPicAllocator::SetMaxTextureSize(SIZE maxTextureSize)
     }
     return S_OK;
 }
-
-STDMETHODIMP CMemSubPicAllocator::SetCurVidRect(RECT curvidrect)
-{
-    m_curvidrect = curvidrect;
-    return __super::SetCurVidRect(curvidrect);
-}
diff --git a/src/SubPic/MemSubPic.h b/src/SubPic/MemSubPic.h
index 903639c02..f078074d0 100644
--- a/src/SubPic/MemSubPic.h
+++ b/src/SubPic/MemSubPic.h
@@ -68,7 +68,6 @@ class CMemSubPicAllocator : public CSubPicAllocatorImpl, public CCritSec
 {
     int m_type;
     CSize m_maxsize;
-    CRect m_curvidrect;
 
     std::vector<std::pair<size_t, BYTE*>> m_freeMemoryChunks;
 
@@ -82,5 +81,4 @@ public:
     void FreeSpdBits(SubPicDesc& spd);
 
     STDMETHODIMP SetMaxTextureSize(SIZE maxTextureSize) override;
-    STDMETHODIMP SetCurVidRect(RECT curvidrect) override;
 };
diff --git a/src/SubPic/SubPicImpl.h b/src/SubPic/SubPicImpl.h
index 5d42902eb..c6dc8aa8e 100644
--- a/src/SubPic/SubPicImpl.h
+++ b/src/SubPic/SubPicImpl.h
@@ -113,7 +113,7 @@ public:
 
 class CSubPicAllocatorImpl : public CUnknown, public ISubPicAllocator
 {
-private:
+protected:
     CCritSec m_staticLock;
     CComPtr<ISubPic> m_pStatic;
 

```

matthewmauger:
@sorayuki: Since @kasper93 has confirmed that this value is unnecessary in the subclass, would you like to remove it as per his diff and resubmit? It's a much cleaner fix than your original change. Well done for identifying the issue!

XhmikosR:
I see @kasper93 has https://github.com/kasper93/mpc-hc/commit/139335118f3a520570ab4d8ab0a4960225c83dfc which should take care of this.

PR #180, opened 2017-04-14
[Suggestion] Rating System, Repeat & Shuffle ICONS

maybe0ne:
Sorry but I don't know how pull request work, I wanted to start a discussion/suggestion post.
Hi, I've got some suggestions for MPC-HC:

1. **Shuffle & Repeat icons**: I know that you can right click the playlist and select Shuffle to enable it but it could be more user friendly just by having the icons in one click. Also if you click another time the repeat button, it should show a 1 (in a circle) which means that the current song will be repeated, not the whole playlist after it ends (if shuffle is not enabled).
2. **Rating system**: It would be awesome to integrate the metadata of songs, RATING MM is the correct metadata used. If you set the RATING MM to 1, it shows 1 star **(see 5 & 6 on the image)**. So my idea is to have the rating right into MPC-HC so we can rate songs and then sort them by rating **(see 3 on the image)** so the most favorited songs are at the top of the list and the worst down there. You can change the rating when you open a folder with songs **(see 5 & 6 on the image)**.

Thanks! What do you think about this suggesiton, is it good?

mpc-star-rating-mockup-shuffle-repeat


XhmikosR:
https://trac.mpc-hc.org/

PR #182, opened 2017-07-19
Fix hash calculation for files with Unicode paths.

clsid2:
This fixes search on OpenSubtitles for such files.

Fixes #6231.

### DIFF
--- a/docs/Changelog.txt
+++ b/docs/Changelog.txt
@@ -14,3 +14,4 @@ next version - not released yet
 * Updated sanear to v0.3
 ! Ticket #2539: Subtitle downloader dialog could be opened on disabled monitor
 ! Fix text subtitle rendering in Avisynth
+! Fix hash calculation for files with Unicode paths. This fixes search on OpenSubtitles for such files.
--- a/src/mpc-hc/SubtitlesProvider.cpp
+++ b/src/mpc-hc/SubtitlesProvider.cpp
@@ -434,7 +434,7 @@ SRESULT SubDB::Hash(SubtitlesInfo& pFileInfo)
     } else {
         CFile file;
         CFileException fileException;
-        if (file.Open(CString(pFileInfo.filePath.c_str()),
+        if (file.Open(CString(pFileInfo.filePathW.c_str()),
                       CFile::modeRead | CFile::osSequentialScan | CFile::shareDenyNone | CFile::typeBinary,
                       &fileException)) {
             file.Read(&buffer[0], PROBE_SIZE);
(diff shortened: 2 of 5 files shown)


osirisgreene:
This looks good to me. I didn't try to actually find subtitles for a file with a unicode pathname, but the code itself checks out.

If I recall correctly (and I only foggily remember), there was some small set of files where it would find non-OST subtitles (from another provider), but not OST? Was this due to this issue --- i.e., is it that these other providers are expecting the hash calculation, however erroneous, from a UTF-8 pathname? I don't remember on that.

But yes, it seems good.

clsid2:
The other providers don't use (or need) a hash to return results.

osirisgreene:
Ok, then it looks good overall. I'd approve.

XhmikosR:
My only concern is if we need a separate unicode filename and if we could just use unicode everywhere without duplicating stuff.

osirisgreene:
What do you mean by "without duplicating stuff"?

XhmikosR:
`filePath` and `filePathW` for example.

osirisgreene:
Oh, I see. Well, it's really a precautionary measure for if things (older
parts of the Win API/MFC API/DirectX API) can't handle it. We know that all
the use cases (except OST) work fine with the current handling, so there is
no reason to break it --- we can try out UTF16 later for all purposes, but
this would take more comprehensive testing.

For now, take the useful bug fix and consider the technical elegance later.

"The perfect is the enemy of the good."

clsid2:
The Win API needs UTF16 and the regex stuff doesn't work with wide strings. I already tried a different approach initially, but it was messy and kept breaking. Current solution is simplest atm.

osirisgreene:
If that's the case, then this is not only a working solution, but also apparently the most elegant one. A proper bug fix.

I motion for the merge.

osirisgreene (approved):
Did not know that the review function was a button on Github. I formalize my approval. (I read the code before I made my previous comments, just didn't see the giant "Review changes" button ... whoops)

kasper93 (approved):

kasper93:
@clsid2: Please add changelog entry about the fix.

XhmikosR:
We should mention the ticket in Changelog too.

kasper93:
And squash it in one commit. Then we can merge.

XhmikosR:
We can squash via the GitHub UI :)

clsid2:
I have got some more patches in main branch that could be cherry-picked.
https://github.com/clsid2/mpc-hc/commits/develop

XhmikosR:
I merged it manually because it had useless stuff in it.

@clsid2: contrib/pre-commit.sh and contrib/run_astyle.bat for the future. Every change that has a ticket should be referenced in Changelog so that we keep track of what's changed.

Thanks for the patch!

PR #183, opened 2017-07-20
Fallback exception handling for when crash reporter is unavailable

clsid2:
This will be used if user removes CrashReporter folder. It is also useful for custom/unofficial builds, for which DrDump is useless.

Additionally I could add an option to the installer to not install CrashReporter. This was requested by users. Opinions welcome.

@XhmikosR, could you give access to DrDump? I would like to redirect some common third party crashes to tickets on Trac, where I explain how user can solve those problems.

If any of the new volunteers have good debugging skills with Visual Studio, perhaps they could attempt to fix any of the top crashes in mpc-hc.

### DIFF
--- a/src/mpc-hc/ExceptionHandler.cpp
+++ b/src/mpc-hc/ExceptionHandler.cpp
@@ -1,0 +1,216 @@
+/*
+ * (C) 2017 see Authors.txt
+ *
+ * This file is part of MPC-HC.
+ *
+ * MPC-HC is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MPC-HC is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "stdafx.h"
+#include "ExceptionHandler.h"
+#include <windows.h>
+#include <psapi.h>
+
+#ifndef _DEBUG
+
+LPCWSTR GetExceptionName(DWORD code)
+{
+    switch (code) {
+        case EXCEPTION_ACCESS_VIOLATION:
+            return _T("ACCESS VIOLATION");
+        case EXCEPTION_ARRAY_BOUNDS_EXCEEDED:
+            return _T("ARRAY BOUNDS EXCEEDED");
+        case EXCEPTION_BREAKPOINT:
+            return _T("BREAKPOINT");
+        case EXCEPTION_DATATYPE_MISALIGNMENT:
+            return _T("DATATYPE MISALIGNMENT");
+        case EXCEPTION_FLT_DENORMAL_OPERAND:
+            return _T("FLT DENORMAL OPERAND");
+        case EXCEPTION_FLT_DIVIDE_BY_ZERO:
+            return _T("FLT DIVIDE BY ZERO");
+        case EXCEPTION_FLT_INEXACT_RESULT:
+            return _T("FLT INEXACT RESULT");
+        case EXCEPTION_FLT_INVALID_OPERATION:
+            return _T("FLT INVALID OPERATION");
+        case EXCEPTION_FLT_OVERFLOW:
+            return _T("FLT OVERFLOW");
+        case EXCEPTION_FLT_STACK_CHECK:
+            return _T("FLT STACK CHECK");
+        case EXCEPTION_FLT_UNDERFLOW:
+            return _T("FLT UNDERFLOW");
+        case EXCEPTION_GUARD_PAGE:
+            return _T("GUARD PAGE");
+        case EXCEPTION_ILLEGAL_INSTRUCTION:
+            return _T("ILLEGAL_INSTRUCTION");
+        case EXCEPTION_IN_PAGE_ERROR:
+            return _T("IN PAGE ERROR");
+        case EXCEPTION_INT_DIVIDE_BY_ZERO:
+            return _T("INT DIVIDE BY ZERO");
+        case EXCEPTION_INT_OVERFLOW:
+            return _T("INT OVERFLOW");
+        case EXCEPTION_INVALID_DISPOSITION:
+            return _T("INVALID DISPOSITION");
+        case EXCEPTION_INVALID_HANDLE:
+            return _T("INVALID HANDLE");
+        case EXCEPTION_NONCONTINUABLE_EXCEPTION:
+            return _T("NONCONTINUABLE EXCEPTION");
+        case EXCEPTION_PRIV_INSTRUCTION:
+            return _T("PRIV INSTRUCTION");
+        case EXCEPTION_SINGLE_STEP:
+            return _T("SINGLE STEP");
+        case EXCEPTION_STACK_OVERFLOW:
+            return _T("STACK OVERFLOW");
+        case 0xE06D7363:
+            return _T("UNDEFINED C++ EXCEPTION");
+        default:
+            return _T("[UNKNOWN]");
+    }
+}
+
+HMODULE GetExceptionModule(LPVOID address, LPWSTR moduleName)
+{
+    HMODULE moduleList[1024];
+    DWORD sizeNeeded = 0;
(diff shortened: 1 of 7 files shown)


kasper93:
We don't need/want minidumps. We use drdump for a reason. We had minidumps for years, no one was reporting crashes, except some sinuous ones. First release after drdump integration we resolved a lot of crashes, and none of them was trivial. And I mean a lot of never seen before top crashers. 

If user want to report crash, without minidump, he can do it manually if needed. But he really should use drdump. I mean it helps a lot. And privacy concerns are exaggerated. 

> This will be used if user removes CrashReporter folder. It is also useful for custom/unofficial builds, for which DrDump is useless.

Minidumps are also useless for custom builds where we don't have symbols and even binary. 

> Additionally I could add an option to the installer to not install CrashReporter. This was requested by users. Opinions welcome.

I have welcome wizard somewhere. It ask few questions about dumps, subtitles, updates.

> could you give access to DrDump? I would like to redirect some common third party crashes to tickets on Trac, where I explain how user can solve those problems.

I can give you access if you send me nick and email of drdump account. 

> If any of the new volunteers have good debugging skills with Visual Studio, perhaps they could attempt to fix any of the top crashes in mpc-hc.

Well to fix we need to have crashes... The majority of remaining crashes in in third-party libraries. We have some crashes, that could be fixed, but there is nothing pressing. And I would wait before giving access everyone to user memory dumps. Even if it is only MPC-HC address space.

clsid2:
This does not generate a minidump! It only displays an error dialog with crash details, most importantly the name of the module in which the crash occurred.

Dialog looks similar to this:
http://i86.servimg.com/u/f86/19/48/78/70/20160611.jpg

I guess I should rename the files. I used some old code as a basis.

DrDump nick: clsid
mail: clsid2@users.sourceforge.net

kasper93:
> This does not generate a minidump! It only displays an error dialog with crash details, most importantly the name of the module in which the crash occurred.

Oh, sorry. I haven't read the code, to see what it does. In this case it is actually good addition.

src/mpc-hc/MiniDump.cpp
```diff
@@ -0,0 +1,196 @@
...
+        case EXCEPTION_NONCONTINUABLE_EXCEPTION:
+            return _T("NONCONTINUABLE EXCEPTION");
+        case EXCEPTION_PRIV_INSTRUCTION:
+            return _T("PRIV INSTRUCTION");
+        case EXCEPTION_SINGLE_STEP:
+            return _T("SINGLE STEP");
+        case EXCEPTION_STACK_OVERFLOW:
+            return _T("STACK OVERFLOW");
+        case 0xE06D7363:
+            return _T("C++ EXCEPTION");
```
kasper93: I would add "Unhandled..."

src/mpc-hc/MiniDump.cpp
```diff
@@ -0,0 +1,196 @@
...
+        default:
+            return _T("[Unknown]");
+    }
+}
+
+HMODULE GetExceptionModule(LPVOID address, LPWSTR moduleName)
+{
+    HMODULE moduleList[1024];
+    DWORD sizeNeeded = 0;
+    if (FALSE == EnumProcessModules(GetCurrentProcess(), moduleList, 1024, &sizeNeeded) || sizeNeeded < sizeof(HMODULE)) {
```
kasper93: This size should be in bytes, not element count.

clsid2:
I have renamed the files. 

Changed to "C++ Exception" tro "Undefined C++ Exception".

src/mpc-hc/ExceptionHandler.cpp
```diff
@@ -0,0 +1,196 @@
...
+        default:
+            return _T("[UNKNOWN]");
+    }
+}
+
+HMODULE GetExceptionModule(LPVOID address, LPWSTR moduleName)
+{
+    HMODULE moduleList[1024];
+    DWORD sizeNeeded = 0;
+    if (FALSE == EnumProcessModules(GetCurrentProcess(), moduleList, 1024, &sizeNeeded) || sizeNeeded < sizeof(HMODULE)) {
```
XhmikosR: Nitpicking but move the check value to the right please.

src/mpc-hc/ExceptionHandler.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
```
XhmikosR, line 21: We need license headers for the new files.

src/mpc-hc/MiniDump.cpp
XhmikosR: Generally IMO we should use the helper macros like `_countof` instead of using hardcoded numbers.

kasper93:
Eh. I haven't submit review comment yesterday.

The idea is good, but we might extend it to drop crash log file. Dump stacktrace to file, it will be easier for users to share than minidumps and give much more info, what has happened. This commit itself is fine, I can take  a look into simple crash report log.

clsid2:
Headers added. Fixed nitpicks.

src/mpc-hc/ExceptionHandler.cpp
```diff
@@ -0,0 +1,216 @@
...
+        default:
+            return _T("[UNKNOWN]");
+    }
+}
+
+HMODULE GetExceptionModule(LPVOID address, LPWSTR moduleName)
+{
+    HMODULE moduleList[1024];
+    DWORD sizeNeeded = 0;
+    if (!EnumProcessModules(GetCurrentProcess(), moduleList, _countof(moduleList), &sizeNeeded) || sizeNeeded < sizeof(HMODULE)) {
```
XhmikosR: Are you sure `sizeNeeded < sizeof(HMODULE)` is right?

src/mpc-hc/ExceptionHandler.cpp
```diff
@@ -0,0 +1,216 @@
...
+            }
+        }
+    }
+
+    if (curModule == -1) {
+        return NULL;
+    }
+
+    if (!GetModuleFileName(moduleList[curModule], moduleName, MAX_PATH)) {
+        return NULL;
```
XhmikosR: Shouldn't we use `nullptr` for pointers?

src/mpc-hc/ExceptionHandler.cpp
```diff
@@ -0,0 +1,216 @@
...
+    if (GetExceptionModule(exceptionInfo->ExceptionRecord->ExceptionAddress, module)) {
+        moduleName = module;
+    } else {
+        moduleName = _T("[UNKNOWN]");
+    }
+
+    uint64_t codeBase = (uint64_t)GetModuleHandle(NULL);
+    uint64_t offset   = (uint64_t)exceptionInfo->ExceptionRecord->ExceptionAddress - codeBase;
+
+    swprintf_s(message, sizeof message / sizeof * message, _T(\
```
XhmikosR: You should probably use `_countof` here too.

XhmikosR:
Thanks for the quick pushes! Now, if @kasper93 confirms the current patches are OK, this would need an astyle run and can be merged with the next batch of patches.

kasper93:
@clsid2 I did quick test. I've thrown c++ exception and it says `C:\WINDOWS\System32\KERNELBASE.dll` while in this case it should say `mpc-hc.exe` :> Can you take a look?

clsid2:
That is normal since that exception is a bit special. The code that raises the exception is in kernelbase. So test with for example a null pointer exception.

Astyle has already been applied a while ago.

@XhmikosR, running Astyle on vanilla clone changes several files here. Only line-endings afaict. I use checkout "as-is" for line-endings.

XhmikosR:
@clsid2: use git config --global core.autocrlf true` since that is what we currently use and try again. I guess eventually we could switch everything to LF but that's for another time.

kasper93:
> That is normal since that exception is a bit special. The code that raises the exception is in kernelbase. So test with for example a null pointer exception.

Well I know, but in this case maybe we shouldn't show module or like I said show `mpc-hc`? Dunno what's best, so I will let you decide. 

> use git config --global core.autocrlf true` since that is what we currently use and try again. I guess eventually we could switch everything to LF but that's for another time.

Line endings are not the problem. CRLF is fine. Problem is that `autocrlf ` doesn't do what we want. It doesn't fix line endings it just convert all LF in repo to CRLF locally. In result it hides inconsistencies in repository and allow to miss them. This is hard topic, that isn't that important as long as everyone uses the same `autocrlf` option. So @clsid2 please set the config as our contributing guidelines says. 

---
Patch LGTM. Further work can be done in other patches if needed. Thanks.

clsid2:
Merged previous patches. LGTM.

It may be any module that triggered it. These kind of crashes need stacktrace.

kasper93 (approved):

kasper93:
Thanks!

PR #186, opened 2017-07-31
Improve subtitle selection

clsid2:
This basically allows
moviename.LANG.srt
to match with any value specified in MPC-HC options, even for non-ISO codes.
Example: https://trac.mpc-hc.org/ticket/4112


kasper93:
@clsid2 I will approve this, but please monitor for any errors and false positives after this change. So we can fix this in cases.

PR #187, opened 2017-08-02
Add setting for custom filename on save image

rekkoha:
This adds an option in advanced settings to allow users to customize the default filename used for snapshots when using File > Save Image (or pressing F5). Currently snapshot filenames are hardcoded like `video.mkv_snapshot_01.12_[2017.08.01_16.59.08]`. This option lets users choose their own filename format string, defaulted to `%f_snapshot_%t_[%Y.%m.%d_%H.%M.%S]` to copy current behavior.

Trac: https://trac.mpc-hc.org/ticket/4241

### DIFF
--- a/src/mpc-hc/MainFrm.cpp
+++ b/src/mpc-hc/MainFrm.cpp
@@ -4883,9 +4883,21 @@ static void ReplaceInvalidFormatSpecifiers(CString& str)
     }
 }
 
-static CString MakeSnapshotFileName(const CString& title, const CString& vidPos = _T(""))
+CString CMainFrame::MakeSnapshotFileName()
 {
     const auto& settings = AfxGetAppSettings();
+
+    CString title;
+    CString vidPos;
+    if (GetPlaybackMode() == PM_FILE) {
+        title = GetFileName();
+        vidPos = GetVidPos();
+    } else if (GetPlaybackMode() == PM_DVD) {
+        title = GetFileName();
+        vidPos = GetVidPos();
+    } else if (GetPlaybackMode() == PM_DIGITAL_CAPTURE) {
+        title = m_pDVBState->sChannelName;
+    }
 
     CString formatString = settings.strSnapshotNameFormat;
     CString escapedTitle = EscapeFormatSpecifiers(title);
@@ -4952,16 +4964,8 @@ void CMainFrame::OnFileSaveImage()
         return;
     }
 
+    CString snapshotName = MakeSnapshotFileName();
     CPath psrc(s.strSnapshotPath);
-
-    CString snapshotName = _T("snapshot");
-    if (GetPlaybackMode() == PM_FILE) {
-        snapshotName = MakeSnapshotFileName(GetFileName(), GetVidPos());
-    } else if (GetPlaybackMode() == PM_DVD) {
-        snapshotName = MakeSnapshotFileName(GetFileName(), GetVidPos());
-    } else if (GetPlaybackMode() == PM_DIGITAL_CAPTURE) {
-        snapshotName = MakeSnapshotFileName(m_pDVBState->sChannelName);
-    }
     psrc.Combine(s.strSnapshotPath, snapshotName);
 
     CSaveImageDialog fd(s.nJpegQuality, nullptr, (LPCTSTR)psrc,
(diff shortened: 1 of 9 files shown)


src/mpc-hc/MainFrm.cpp
```diff
@@ -4948,17 +5022,17 @@ void CMainFrame::OnFileSaveImageAuto()
...
     }
 
-    CStringW prefix = _T("snapshot");
+    CString snapshotName = _T("snapshot");
     if (GetPlaybackMode() == PM_FILE) {
-        prefix.Format(_T("%s_snapshot_%s"), GetFileName(), GetVidPos());
+        snapshotName = MakeSnapshotFileName(GetFileName(), GetVidPos());
     } else if (GetPlaybackMode() == PM_DVD) {
-        prefix.Format(_T("dvd_snapshot_%s"), GetVidPos());
+        snapshotName = MakeSnapshotFileName(GetFileName(), GetVidPos());
```
clsid2: I don't think that GetFileName() is valid for DVDs.

src/mpc-hc/MainFrm.cpp
```diff
@@ -4880,15 +4954,15 @@ void CMainFrame::OnFileSaveImage()
 
     CPath psrc(s.strSnapshotPath);
 
-    CStringW prefix = _T("snapshot");
+    CString snapshotName = _T("snapshot");
```
clsid2: This assignment can be move to end of if/else for unhandled GetPlaybackMode() values.

src/mpc-hc/MainFrm.cpp
```diff
@@ -4948,17 +5022,17 @@ void CMainFrame::OnFileSaveImageAuto()
         return;
     }
 
-    CStringW prefix = _T("snapshot");
+    CString snapshotName = _T("snapshot");
```
clsid2: This code block is similar as the other one. Could be refactored or moved into MakeSnapshotFileName().

src/mpc-hc/mpc-hc.rc
```diff
@@ -3652,23 +3677,8 @@ END
...
-    IDS_CMD_HELP            "/help /h /?\tShow help about command line switches"
-    IDS_PPAGEADVANCED_SCORE "Threshold value for subtitles score which are going to be automatically downloaded. Higher values means that more accurately matched subtitles will be loaded, lower values could result in incorrect subtitles being loaded, but there is no one perfect value. Pick one that works best for you."
-    IDS_PPAGE_FS_CLN_AUDIO_DELAY "Audio Delay (ms)"
-    IDS_PPAGEADVANCED_DEFAULTTOOLBARSIZE 
-                            "Size in pixels of the default toolbar."
-    IDS_PPAGEADVANCED_USE_LEGACY_TOOLBAR 
-                            "Use legacy toolbar instead of new vectorized one."
-    IDS_SUBMENU_COPYURL     "Copy URL"
+    IDS_PPAGEADVANCED_SNAPSHOT_NAME_FORMAT 
+                            "Filename format used for snapshots on ""Save Image"". %f is the video filename, %t is the current video position, %Y, %m, %d is the current year, month, day, etc."
```
clsid2: "Filename template that is used for"

For example: "%f_snapshot_%t_[%Y.%m.%d_%H.%M.%S]", where %f equals video file name, %t current video timestamp, %Y current year, etc.

kasper93:
Thanks for patch, but I don't think this is best approach. 95% of people will not understand what to type there and how this work. And it is not good to give user full control over format string. I would prefer few preset that every user can understand and select best one. Similar to what @clsid2 done in his patch. But I'm open for discussion here.

rekkoha:
Is the main concern the possibility of a security vulnerability? I believe it is better to give users the option of having a custom format string, since power users will most likely be the ones using this feature. FWIW, VLC is doing the same thing: https://wiki.videolan.org/Documentation:Play_HowTo/Format_String/

clsid2:
This feature should also be useful for less experienced PC users. So perhaps my simpler approach would indeed be better. Which custom string would you personally use if your patch would be used?

rekkoha:
I don't think less experienced users would use this feature. The users who want to change the filename prefer full control over it, like the person who created the original Trac ticket for this feature: https://trac.mpc-hc.org/ticket/4241

If less experienced users that are interested in this feature do exist, I think they will be ok as long as we provide sufficient documentation. It seems to be working for VLC.

Personally I prefer the video time at the beginning, so I would use `%t_%f_%Y%m%d_%H%M%S`

PR #190, opened 2017-08-06
Update Compilation.md.

anthnich:
Following _Compilation.md_, MPC-HC was unable to compile, with many missing includes, in Windows 10/Visual Studio 2017 until I selected **Windows Universal CRT SDK** under "Individual Components -> Compilers, build tools, and runtimes". Selecting and ensuring that **Windows Universal CRT SDK** allowed compilation without issues.

### DIFF
--- a/docs/Compilation.md
+++ b/docs/Compilation.md
@@ -12,7 +12,7 @@ Choose `Use Git from the Windows command prompt`. This isn't mandatory, so if yo
 ### Visual Studio 2017 or 2015
 
 1. Install Visual C++, part of Visual Studio (any edition will work fine).
-   Make sure to select **Windows 8.1 SDK** and **MFC and ATL support** during installation.
+   Make sure to select **Windows 8.1 SDK**, **MFC and ATL support**, and **Windows Universal CRT SDK** during installation.
 2. Make sure you have installed all available updates from Microsoft Update
 3. Install the DirectX SDK (June 2010) → <https://go.microsoft.com/fwlink/?LinkID=71193>
 


XhmikosR:
IMHO your installation was wrong from the beginning. I had no issue just doing what the current docs say, that is why I documented it.

anthnich:
Followed the guide to the letter. Got a lot of missing header files in the third party libraries for mpc-hc. Did some googling regarding missing headers, where it lead me to a recommendation to install Windows Universal CRT SDK. Modified the VS2017 to include that. Compiled fine after that.

anthnich:
Followed the guide to the letter and that was my outcome. 

Also, I needed to modify my install and include "Windows XP support for C++" for LAVFILTERS to compile in the solution.

XhmikosR:
XP is no longer supported.

As for the other change, I definitely didn't need to do it myself.

anthnich:
I know. 

So I can look into it more to see what possibly got messed up on my end, are you running Windows 10 and VS2017?

clsid2:
Are you sure? build_lavfilters.bat does not seem to override the toolchain and SDK, so it uses the default ones as specified in the LAV projects files (which currently is XP toolchain).

This need to be added to /property parameter for VS2017:
PlatformToolset=v141;WindowsTargetPlatformVersion=8.1

XhmikosR:
@anthnich: yes, on Windows 10, VS2017 and only C++ with those 2 extra options.

@clsid2: not sure I follow? It might be possible we have missed to change the toolset and stuff in our LAV fork if you mean that.

clsid2:
Yes, LAV wasn't updated.

XhmikosR:
Ok that seems a separate issue. /CC @kasper93 for updating our lav fork to use vs2017 with the SDK we pass and target >= Vista.

XhmikosR:
@kasper93: can we close this?

kasper93:
Indeed doing dry test it doesn't seem to auto select Universal CRT. Which is weird, but I didn't actually start the installation. 

Maybe someone could confirm that we need to select it explicitly? And if anything else is missing.

XhmikosR:
I honestly didn't need to explicitly select it when I cleaned installed VS2017 a month or so ago.

I did choose C++ environment though.

kasper93:
@anthnich reported that our current guide dosn't work, maybe we should merge it? It doesn't harm to have one more component listed and if this help people to setup environment easier, why not?

XhmikosR:
I'll do a test tomorrow on my VM and report back.

XhmikosR:
Indeed, it's not selected. I don't remember enabling it manually but that's what I must have done.

@kasper93: should we switch to the default SDK that ships with VS2017 or still use 8.1? The 8.1 SDK is 1.4GB more.

kasper93:
We should use `8.1` as it is the one to be used to Vista+ targets. There are no issues with using Windows 10 SDK, but since we don't really need it we can keep it with `8.1`.

PR #191, opened 2017-08-14
Adjust ISOLang of chinese.

kongfl888:
Adjust ISOLang of chinese.


kasper93:
What's wrong with `Chinese (simplified)`? I mean now there is ambiguity because there are two "Chinese" so "lang to iso" function will return first one, but this is implementation detail. It shouldn't be ambiguous.

kongfl888:
Both "chi" and "zho" represent Chinese which is a macro concept in the ISO 639-2 Code.  And there is no simplified Chinese in languages, I mean speaking, not writing to the text.
("chi" is more commonly used and recognized by users. If "lang to iso" return first, that is, chi, I think this can be accepted.)

clsid2:
These values are primarily used for subtitles, so written text.

kongfl888:
@clsid2 Thank you. But "chi" is still an acronym for "Chinese", "zho" too, and "Chinese" is a macro language, and you can't limit it to one of them. There are more than one language in my country. 
These are generally accepted, whether in iso639 or in use. I don't think there's anything to discuss about that. 
In this case, these short words are based on iso639, I think we should not modify them according to the personal meaning. So that mistakes left before should be fixed.

kasper93:
Ok, if you feel it is better I will merge it. Thanks.

kongfl888:
Thank you.

PR #193, opened 2017-09-09
Shorten tooltip width in Internal Filters

phineas12:
"IDS_INTERNAL_LAVF_WMV" is missing line breaks
https://trac.mpc-hc.org/ticket/5905#comment:4


kasper93:
Looks good, thanks.

PR #196, opened 2017-09-27
Youtube-dl integration

nparkanyi:
This enables streaming Youtube videos with youtube-dl by giving a Youtube URL in the Open File dialogue or on the command line. Supports individual videos and playlists. Requires youtube-dl to be present in PATH variable.

PR #198, opened 2017-12-01
Audio extract -- extract audo clip to MP3 file while playing media file

kashwy:
when playing a media file, user can click "Grab Audio" from the context menu by right click the mouse(the menu item will be checked), then the player will start to record audio as MP3 file to the personal music folder, until user unchecked the menu item or the media file stopped playing.
If the menu item has been checked before playing, the when a media file begin to play, the audio will be recorded automatically.
@kasper93


ExTee:
@kasper93

src/DSUtil/DSUtil.cpp
```diff
@@ -159,7 +159,8 @@ bool IsAudioWaveRenderer(IBaseFilter* pBF)
            clsid == __uuidof(CNullAudioRenderer) ||
            clsid == __uuidof(CNullUAudioRenderer) ||
            clsid == CLSID_SANEAR_INTERNAL ||
-           clsid == CLSID_SANEAR;
+           clsid == CLSID_SANEAR ||
+		   clsid == CLSID_InfTee; // allow the inftee to be connected from Audio Switch , 171003
```
danilaml: inconsistent indent

kasper93:
What are the purpose of this functionality? There are better ways to extract audio form medial file. But I can see this useful as quick grab for audio part quickly without any other software.

I might be missing the point, because I'm not sure how exactly this functionality needs to be used. But first thing that comes to mind is to separate whole filter and compression from player. It will be more flexible to have external filter that does dirty job and just have integration code in the player to request recording (filter connection etc.). 

Frankly I think the good place for such functionality would be audio decoder which have access to original stream which would allow to skip recompression in some cases.

kashwy:
@kasper93 
Advantages of this feature:
1. Video courses from online educational websites such as Coursera can be quickly converted to an MP3 version, which can be imported into an MP3 player, allowing for quick and portable review of lecture material while walking,driving or in transit. The user is able to convert the video into MP3 while they are watching the video, and does not need an extra step, saving time.

2. Ease of obtaining an MP3 copy of a music video WHILE watching it, without the need to look for another tool. After the user finishes watching the video, the MP3 file is already generated.

Using audio decoder for this task requires various formats of the source. Our solution handles any file format.
Re-compression is needed for users who wish to use different MP3 bit rate.  
We included an options page just like in our last commit.
We need to process this in the player because the player is more powerful than the external filter. This provides easier buffer and thread handling, etc.

Please consider the changes,
Thanks.

chros73:
Hi! Take a look at this [comment](https://github.com/mpc-hc/mpc-hc/pull/196#issuecomment-359254041) ...

PR #199, opened 2018-02-07
HdmvClipInfo: Support UHD Blu-ray discs

Nevcairiel:


frymezim:
I think MPC-HC will no longer maintained, https://forum.doom9.org/showthread.php?t=175209
Instead @clsid2 created his own repo it would be better to post pull request there.
https://github.com/clsid2/mpc-hc/

PR #203, opened 2018-09-26
Cast on Smart Screen Bug

adeljaffan:
at first all thanks for this awesome program, i have two issues 
1- why the bug report page not user friendly and so complicated , honestly i didn't know how to send bug report !!
2- my main issue : 
i have LG TV 43" Smart Screen , Lenovo Laptop Y520 and latest version of mpc 
when i try to run any video format and cast on the screen , every thing is ok
but after a few minutes the laptop freezing and screen stop casting , keyboard doesn't working
  and that let me restart the laptop to stop freezing
when i use vlc program , every thing ok !
all thanks and respect


adeljaffan:
at first all thanks for this awesome program, i have two issues
1- why the bug report page not user friendly and so complicated , honestly i didn't know how to send bug report !!
2- my main issue :
i have LG TV 43" Smart Screen , Lenovo Laptop Y520 and latest version of mpc
when i try to run any video format and cast on the screen , every thing is ok
but after a few minutes the laptop freezing and screen stop casting , keyboard doesn't working
and that let me restart the laptop to stop freezing
when i use vlc program , every thing ok !
all thanks and respect

vBm:
This is not the place for bug reports.

PR #205, opened 2019-06-09
Make player.html video more responsive

stdedos:
Until some proper technology is used, this change would make the video a bit more responsive.

The change is small enough, so that browser CPU utilization is kept below 7% on average (while the normal is about 3-4% on i5-4430). 50ms delay would make the video more stable, but CPU usage would be double.

### DIFF
--- a/src/mpc-hc/res/web/javascript.js
+++ b/src/mpc-hc/res/web/javascript.js
@@ -489,12 +489,12 @@ function loadSnapshot() {
 
 function onLoadSnapshot() {
     "use strict";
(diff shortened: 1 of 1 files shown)


MarkusGaugg:
This fork https://github.com/clsid2/mpc-hc/ is active in maintenance mode. Please post your Pull Request here, because this repository is abandoned by devs.

stdedos:
@MarkusGaugg I am not sure I feel confident enough that that's going to help.

Regardless, here it is: https://github.com/clsid2/mpc-hc/pull/249

PR #207, opened 2019-09-08
Steroscopic 3D video and images (Specifically, Side by Side) usually …

rtrded:
Steroscopic 3D video and images (Specifically, Side by Side) usually need to be edited in some software before they can be presented on a 3D TV as the left and right halves of the image must be "squshed" horisontally in place (so a large black column appears in the middle and two smaller columns on the left and right edge).

Additionally, fine adjustments to the seperation and the left/right eye elevation need to be made to correct for lens calibration issues with the source recording.
Currently there is no playback software which allows the user to make these adjustments to the source video in real time.
This feature provides all this functionality allowing raw 3D SBS content to be viewed on a TV without the need of recompression and time consuing video/image editing

The keys for controlling all this are within the hot key screen at the bottom and are currently mapped in a similar form to the WASD movement but
using the IJKL keys to avoid conflicing with more critical keys.

I and K shift the left eye up/down and the right eye down/up
J and L shift the eyes apart and closer together.
The top left and right of these keys are the U and O keys which shrink and enlarge the left and right images.
Over to the left, the 'S' key enables and disables the horisontal Squish of each eye. This is default to enabled.
The purpose is in case someone has pre-squished left and right eyes but wants to adjust the seperation/elevation.

To enable/disable all this 3D functionality, simply press the '3' key.

A sample 3D SBS image is included for testing.
SBS_3D_Sample_Image

Note: It works best with video because images only draw once so changes made to the 3D settings only update on screen if the window is resized.

Note: There was an error when I was trying to compile the latest source before doing this work in MainFrm.cpp so I did a hacky fix (type casting  CComQIPtr<IBaseFilter> pBF = (CComQIPtr<IBaseFilter>)(bIsSplitter ? (CComQIPtr<IBaseFilter>) pSS : (CComQIPtr<IBaseFilter>)pObject); )


rtrded:
MainFrm.h got reverted somehow so that file's changes are missing. Just rewriting that code and will then do a commit.

rtrded:
Removed the sample image from the repo.
Improved the shifting code in the renderer so it scales correctly when the window resizes.
Adding the missing MainFrm.h file changes.
Removed the Next Subtitle key binding so it doesn't clash with the Squish key. (yes, controversial, happy to use a different key, just doing this for now to fix the clash)

rtrded:
Found an issue during testing - only right eye shows if 10bit output is enabled.

JonasNo:
This fork https://github.com/clsid2/mpc-hc/ is active in maintenance mode. Maybe you have better luck there since you have done work. When you have gotten the bugs fixed you **might** get lucky.

